PR #14, opened 2021-03-06
Portfolio Mette Wickert

ettem:
This week I feel fragmented. Combining building my portfolio taking into consideration accessibility requirements and become aquatinted with unit testing has been..not the greatest experience. I am ok with the portfoliobuilding itself, but I am so lost when it comes to how to get it accessible. I have followed the assignment directive thoroughly but have not succeeded on step 6, 7, 8 and 10 on this list: https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Accessibility 

Screenreader test is bananas and so bad, but testing Voice Over on my new MacBook gives a better result. I guess it's a new better version of Voice Over, it can at least read my paragraphs whilst my old MacBooks Voice Over doesn't. I tried adding ARIA attributes but it didn't work, I find it quite difficult and don't know where to start, so getting a pointer on where I could start would make me very happy. Also, I decided to keep my pop-up menu with javascript even though I shouldn't from an accessibility perspective. A question is how the hamburger menu can be so commonly used on mobile devices when you are not supposed to use hidden elements and as little javascript as possible building accessible websites? Am I missing something?

https://quirky-engelbart-0570dc.netlify.app/


about.html
```diff
@@ -0,0 +1,80 @@
...
+        <div class="wrapper-img">
+            <img class="logo" src="./logga_personal.png" alt="logotype" />
+        </div>
+
+        <!--menu desktop and tablet-->
+        <nav class="nav">
+            <a class="nav-item" href="./about.html" >about</a>
+            <a class="nav-item" href="./work.html">work</a>
+            <a class="nav-item" href="./contact.html">contact</a>
+        </nav>
```
KarolCislowski, line 27: your navigation is changing its height when you hover over links. You should reduce your bottom padding when you are showing the border to counter this effect. 


Nav is the only place where you are using the Serif font. it feels odd to me.

about.html
```diff
@@ -0,0 +1,80 @@
...
+
+    <footer class="footer">
+        <div class="copyright">
+            <p>&#169; 2021 Mette Wickert</p>
+        </div>
+        <div class="footer-icons">
+            <a class="icon1-footer" href="https://www.linkedin.com/in/mettewickert"><img class="footer-icon" src="linkedin_icon.png" alt="linkedinicon" /></a>
+            <a class="icon2-footer" href="https://github.com/ettem"><img class="footer-icon" src="github_icon.png" alt="githubicon" /></a>
+        </div>
+    </footer>
```
KarolCislowski: Your footer is not on the bottom of the page and it looks unprofessional 😿

contact.html
```diff
@@ -0,0 +1,69 @@
...
+        <nav class="nav">
+            <a class="nav-item" href="./about.html" >about</a>
+            <a class="nav-item" href="./work.html">work</a>
+            <a class="nav-item" href="./contact.html">contact</a>
+        </nav>
+    
+        <!--menu mobile-->
+        <div class="wrapper-mobile-menu">
+            <div></div>
+            <div></div>
```
KarolCislowski: what are these empty divs for? you have a lot of them in your code.

about.html
```diff
@@ -0,0 +1,80 @@
...
+                <p> Learning designer and digital producer located outside Stockholm. 
+                Working with text- and mediaproduction with the intention to create great learning experiences,  
+                good design is of great importance. I strive to develop those skills with studies in digital design at University of Kristianstad. 
+                I also have a new found love for coding and how it nurtures my creativity. Hence I am also adding some front end development skills by participating in the Technigo bootcamp this spring.</p>
+                <p>My future goal is to create greatness by combining my three <span class="about-heart">&#9829;</span> learning, design and code.</p>
+                <p><span class="headline-skills">Tech skills: </span> Html, Css, Javascript, Github</p>
+                <p><span class="headline-skills">Toolbox: </span> Articulate 360, Adobe Creative Cloud</p>
+            </div>
+            <div class="wrapper-imgmyself">
+                <img class="about-img" src="about.png" alt="self-portrait-photo" />
```
KarolCislowski: This image is 9585 × 9586 pixels and over 4MB 🙀 it is way too big and it is loading too long what will make a very bad first impression. compress your pictures in photoshop or use apps like https://tinypng.com/

KarolCislowski (changes requested):
Your page looks nice, but your code needs some cleaning. 

Please fix the issues I pointed in the comments and it will be a very nice portfolio to show to the recruiters.

PR #22, opened 2021-03-07
Technigo Week 8 project: Portfolio

idasylven:
Overall I thought this project was a lot of fun! I love design processes and concept development in general, but I have realised I also seem to really enjoy trying to implement a set design. 

I am however, still struggling a bit with flexboxes. Some sections of the site are not yet right, nor displaying the way I would like them to throughout the media quaries. I am particularly interested on how I achieve so that the content of the page stops expanding and just stays centered for really big screens. I have googled a lot but haven´t been find a solution that works out for me in my preferred manner. 

I find the accessibility topic very interesting! It has been a great opportunity to really think though the HTML semantics, color contrast etc. It was a good practice to try out different browsers for the deployed project, as well as on the physical devices I´ve had access to. I´ll for sure keep working with the validation tools from here on. The VoiceOver tool I found a bit complex to use, but I guess that takes some practice too. I would like to dig deeper into the ARIA attributes when I get the chance, and apply them on a more detailed level.   

Please see below link for deployed project:

https://idasylven-portfolio.netlify.app/


index.html
```diff
@@ -0,0 +1,464 @@
...
+                    </div>
+                </section>
+            </section>
+
+        </article>
+
+        <!-- Other projects
+        (TBC: Revise with future projects) -->
+
+        <article class="more-projects" tabindex="0">
```
KarolCislowski: This section should take the full width of the container. 🎨

index.html
```diff
@@ -0,0 +1,464 @@
...
+
+    <main class="main-container">
+
+        <!-- SUMMARY SECTION 
+            (TBC: Short & personal text of what I have been doing in the past
+            and want to do in the future)-->
+
+        <article class="summary-section" tabindex="0">
+            <h3>
+                <img class="start-figure" src="./start-figure.png" alt="Start figure"> Hello! Lorem ipsum dolor sit amet consectetur, adipisicing elit. Aliquid id sunt cupiditate nulla esse similique amet modi officia, aspernatur quo velit laudantium,
```
KarolCislowski: This section is about you, so write a few words about yourself.

index.html
```diff
@@ -0,0 +1,464 @@
...
+                        A responsive fashion news site using <br> HTML and CSS.
+                    </p>
+                    <div class="tags">
+                        <p class="html-tag">
+                            HTML5
+                        </p>
+                        <p class="css-tag">
+                            CSS3
+                        </p>
+                    </div>
```
KarolCislowski: Font size in your tags is so small that it hard to read 😿

KarolCislowski (changes requested):
It is a very nice portfolio page 🦄  
I left below a shortlist of fixes needed to fit it better to the guidelines. Please make them soon 😸

KarolCislowski (changes requested):
Every section of your website has a different width. It is not a proper layout image

idasylven:
Thank you for your comments! 💫    
I have re-done my portfolio using React, revised the content and layout should be improved too 🤞🏻

KarolCislowski (approved):
Good work with react 🥳

PR #25, opened 2021-03-07
Portfolio-Marielle Ohlsson

Marielle8:
I aimed for the blue level in this project. I tried to style my page something like the figma example even if I haven't signed up for the career changing program. I struggled with flexbox this time and mixed everything up, I still need to fix my skill section, for more section, featured projects section and contact section when it comes to positioning of the items.


index.html
```diff
@@ -0,0 +1,179 @@
...
+        <div class="icon-section">
+          <a class="circle" href="https://github.com/">
+            <img class ="icon" src="assets/icons/githubVit.svg" alt="github-icon"/>
+          </a>
+          <a class="circle" href="https://www.linkedin.com">
+            <img class ="icon" src="assets/icons/linkedinVit.svg" alt="linkedIn-icon"/>
+          </a>
+          <a class="circle" href="https://stackoverflow.com">
+            <img class ="icon" src="assets/icons/stackVit.svg" alt="stack-overflow-icon"/>
+          </a>
```
KarolCislowski: It would be nice to add some hover effects to make your visitor subconscious want to click this link 🧙🏻

index.html
```diff
@@ -0,0 +1,179 @@
...
+    </header>
+  
+    <!--1. About me-->  
+    <main>
+      <section class="about-me-container">
+        <div class="text-about-me" 
+             tabindex="0" 
+             aria-label="information about me"
+        >
+          <p tabindex="0">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
```
KarolCislowski: This section is about you, so write a few words about yourself.

index.html
```diff
@@ -0,0 +1,179 @@
...
+           </p>
+        </div>
+      </section>
+    <!--featured projects with a link-->
+      <section class="featured-projects-container">
+        <h2 class="h2-header"><span class="marker" tabindex="0">FEATURED PROJECTS</span></h2>
+        <div class="link-cards">
+        <div class="project-container">
+          <a class="link-container" href="https://github.com/Marielle8/project-news-site">
+            <img class="project-picture" src="./assets/images/markus-winkler-aId-xYRTlEc-unsplash.jpg" alt="Picture of a typewriter">
```
KarolCislowski: Your project images should be of the same size.

index.html
```diff
@@ -0,0 +1,179 @@
...
+           <p tabindex="0"  >HTML, CSS,Flexbox, Javascript, ES6, JSX, React, React Hooks, Redux, Node.js, Mongo DB, Web Accessibly, API:s, mob-programming, pair-programing, Github
+           </p>
+        </div>
+      </section>
+    <!--featured projects with a link-->
+      <section class="featured-projects-container">
+        <h2 class="h2-header"><span class="marker" tabindex="0">FEATURED PROJECTS</span></h2>
+        <div class="link-cards">
+        <div class="project-container">
+          <a class="link-container" href="https://github.com/Marielle8/project-news-site">
```
KarolCislowski: On the desktop, there should be only 2 projects in the row.

index.html
```diff
@@ -0,0 +1,179 @@
...
+            <img class="project-picture" src="./assets/images/markus-winkler-aId-xYRTlEc-unsplash.jpg" alt="Picture of a typewriter">
+            <h3 class="h3-header">NEWS-SITE</h3>
+            <p class="featured-project-text">Text about the project</p>
```
KarolCislowski: You need to add tags here.

index.html
```diff
@@ -0,0 +1,179 @@
...
+          </a>
+        </div>
+        <div class="project-container">
+          <a class="link-container" href="">
+            <img class="project-picture" src="./assets/images/bootcamp.jpg" alt="Picture of a running person">
+            <h3 class="h3-header">BUSINESS SITE</h3>
+            <p class="featured-project-text">Text about the project</p>
+          </a>
+        </div>
+        </div>
```
KarolCislowski, line 152: You need to add the "other projects" section here.

index.html
```diff
@@ -0,0 +1,179 @@
...
+      </section>
+      <!--My thoughts about tech-->
+      <section class="my-thoughts-container">
+        <h2 tabindex="0" class="h2-header"><span class="marker">My THOUGHTS</span></h2>
+        <div class="my-thoughts-text">
+          <p tabindex="0">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur</p>
+        </div>
+      </section>
```
KarolCislowski: this section shouldn't be a single block of text.

KarolCislowski (changes requested):
Your page looks very nice, but it is a little bit too far from the design provided to you by us. 
Please fix the issues I pointed in the comments and it will be a very nice portfolio to show to the recruiters.

KarolCislowski (changes requested):
You need to centralize some sections 😸  image

index.html
Marielle8: I know I don't have My thoughts any more on my site but I'm not in the career coaching program, I just used this design as inspiration. So I wonder if it's ok that I don't have that section in my project?

KarolCislowski (approved):
👍🏻

PR #30, opened 2021-03-07
Maidelin Rubio PortFolio

rubmaigut:
This week was totally messy for me. I was really sick and I couldn't be active, I  started to work on this project on Friday that I started to recover and I felt much better. 

I feel that this week I couldn't reach the goals, but I will try to fix this project and improve the labs.


KarolCislowski (changes requested):
Your design looks nice, but you need to experiment some more with it. This is what happens when you change your height to pretty small values: 

image

rubmaigut:
> Your design looks nice, but you need to experiment some more with it. This is what happens when you change your height to pretty small values:
> 
> image

Hi, @KarolCislowski I understand your comments, but the site is not designed to be responsive by the height of the browser. I am working with statics sections and I tried to cover the view on different devices but it needs to has a min-height of 708 px. To change this I need to remove the scroll down by section logic. Please let me know what you suggest

KarolCislowski (approved):

PR #34, opened 2021-03-07
portfolio / Anna-Kaisa 

honkoan:
This week I had lots of problems with schedules. The portfolio doesn’t work the way I would like, but I didn’t have time to make it better. I had problems with the flexbox and it took way too much time.

https://silly-pasteur-4234a5.netlify.app/


index.html
```diff
@@ -0,0 +1,225 @@
...
+            <div class="social-media-icons">
+                <a href="https://www.linkedin.com/in/akhonkonen/">
+                    <i class="fab fa-linkedin-in"></i>
+                </a>
+                <a href="https://github.com/honkoan">
+                    <i class="fab fa-github"></i>
+                </a>
+                <a href="https://stackoverflow.com/users/14964436/honkoan?tab=profile">
+                    <i class="fab fa-stack-overflow"></i>
+                </a>
```
KarolCislowski: It would be nice to add some hover effects to make your visitor subconscious want to click this link 🧙🏻

KarolCislowski (approved):
It is a very nice portfolio page 🦄
I left a small tip below to make it even better. 🦄

PR #60, opened 2021-10-08
Portfolio Rebecca Blixt

rebeccablixt:
Link to deployed version:
https://rebecca-blixt-portfolio.netlify.app/

Things I'd like to change in near future: 
I need to complete the paragraphs for some of the sections. I'll probably change the hero image and the profile image when I find better images. Apart from that I'm proud of what I have accomplished!


index.html
```diff
@@ -0,0 +1,402 @@
...
+  <body>
+    <header class="header">
+      <div class="hero-image">
+        <div class="icon-box">
+          <a
+            aria-label="linkedin profile"
+            href="https://www.linkedin.com/in/rebecca-blixt-069272159/"
+            target="_blank"
+          >
+            <svg
```
KaraHowes, line 49: Nice use of the aria-label for accessibility and using target="_blank" so that the user can navigate easily back to your portfolio.

index.html
```diff
@@ -0,0 +1,402 @@
...
+      <!-- Tech -->
+      <section class="part-section">
+        <div class="container">
+          <!-- Tech-section -->
+          <h1>Tech</h1>
+          <p class="paragraph">
+            HTML, CSS, Flexbox, Grid, Javascript, ES6, React, React Hooks,
+            Redux, React Native, Node.js, Web Accessibility, API:s,
+            mob-programming, pair-programming, Github
+          </p>
```
KaraHowes, line 155: I like the use of container and paragraph as classes for consistency

index.html
```diff
@@ -0,0 +1,402 @@
...
+      <section class="part-section">
+        <div class="container">
+          <h1>Featured Projects</h1>
+          <section class="featured-projects">
+            <article class="projects">
+              <div>
+                <a
+                  href="https://github.com/BritishSwede/project-guess-who"
+                  target="_blank"
+                >
```
KaraHowes, line 172: Do you need a label for links (like an aria-label) so that the user knows to where they are navigating? This is more of a discussion comment because I thought they were necessary, but I really am not sure!

index.html
```diff
@@ -0,0 +1,402 @@
...
+                  <img
+                    src="./images/guess-who-image.png"
+                    alt="Image of Guess Who project"
+                  />
```
KaraHowes, line 176: I was advised against using "image of" as an alt description (I think when I performed an HTML validation) because screen-readers already say "an image of".... again, not entirely sure I am correct ;-)

index.html
```diff
@@ -0,0 +1,402 @@
...
+          <h1>Skills</h1>
+          <div class="skills">
+            <article class="">
+              <h2>code</h2>
+              <p class="paragraph">
+                HTML5 <br />
+                CSS3 <br />
+                Javascript ES6 <br />
+                GitHub
+              </p>
```
KaraHowes: It might be easier to add these skills as an unordered list (ul). Right now, it shows fine in the browser, but things might become skewed when you add skills to the toolbox section.

style.css
```diff
@@ -0,0 +1,339 @@
...
+  top: 380px;
+  width: 150px;
+  height: 150px;
+
+  /*  Rounded corners */
+  border-radius: 50%;
+  border: 3px solid whitesmoke;
+  object-fit: cover;
+}
+
```
KaraHowes, line 122: I really like how you break up the code to describe exactly what you are styling and why!

KaraHowes (commented):
What a great Portfolio, Rebecca! You should be proud. It looks extremely professional. 

As far as I can tell, your project meets all of the blue requirements; it can be viewed on the browsers specified and passes HTML and CSS validation. It is responsive and has been constructed using CSS Grid. I admire the consistency in your code and it is well-described throughout.

Congratulations on an excellent project!

index.html
rebeccablixt: Whoops, I missed adding them to the project... Thanks for pointing it out 😄

index.html
rebeccablixt: Thank you for letting me know! Maks told me as well during 1-on-1 so will change that 👍🏼

index.html
rebeccablixt: Didn't think about that! Thank you!

PR #62, opened 2021-10-08
Mariia-Snopok-Portfolio

MarySnopok:
What went well : 
The layout and responsive design. 
Limitations : 
All accessibility validators dont like my colour contrast. I used css variables thus  
colour can be changed easily to match any requirements . I will leave this pallet cause it reflects me more.
Not sure about: 
I will be happy to get a feedback on site accessibility analysis and understand deeper ,what kind of  limitations are present in the current code.
Challenges : 
It was challenging trying to follow the design in figma and interesting . However  it made me a bit confused at the start what affected the code . So If I start from scratch I would start from the html structure and then go further with styling (as Max had suggested on the lecture) .
(http://www.mary-snopok.com)


code/canvas.js
```diff
@@ -0,0 +1,108 @@
+// Inspired by Dribble.
```
madeleinesvensson, line 1: This is seriously sooo cool feature! good job on finding this and getting it to work! really makes the header pop out! also live that you can toggle between colors!

code/index.html
```diff
@@ -0,0 +1,427 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
```
madeleinesvensson, line 4: ohh you added the tags for charing aswell! good job!

code/index.html
```diff
@@ -0,0 +1,427 @@
...
+                <a href="mailto:marysnopok8@gmail.com" role="button" tabindex="0" aria-label="email me" target="_blank"
+                    rel="noreferrer">
+                    <div class="circle-icon">
+                        <img class="icon email" src="https://svg-clipart.com/svg/white/M5r4etM-mail-icon-vector.svg"
+                            alt="email icon">
+                    </div>
+                </a>
+            </li>
+        </ul>
+        <div aria-label="picture with coloured dots" role="complementary" id="dots" class="dots">
```
madeleinesvensson: this is the background?

code/index.html
```diff
@@ -0,0 +1,427 @@
...
+    <main>
+        <!-- *** welcome message *** -->
+        <section class="full-width-text-container align-self-center">
+            <p class="content">
+                <span aria-hidden="true" class="marker"><span class="line"></span>
+                    <span aria-hidden="true" class="point"></span>
+                </span>
+                <span aria-hidden="true" class="hi">
+                    &lt;hi&gt;
+                </span>
```
madeleinesvensson, line 86: really nice the <hi> tag! makes it looks so nice!

code/index.html
```diff
@@ -0,0 +1,427 @@
...
+                I'm an aspiring front-end developer with experience in people operations within IT industry <6. For me
```
madeleinesvensson: is the <6 suposed to be there or is that a tag that is weird? or am I not getting what it means maybe :P?

code/index.html
```diff
@@ -0,0 +1,427 @@
...
+                    &lt;/hi&gt;</span>
+            </p>
+        </section>
+        <!-- *** tech *** -->
+        <section class="align-center">
+            <article class="full-width-text-container">
+                <div class="headings-container">
+                    <h2 class="section-heading tech-heading">tech</h2>
+                </div>
+                <p class="content tech-skills"> HTML, CSS, Flexbox, <span class="stand-out-skills">JavaScript</span>,
```
madeleinesvensson: like the naming of your classes! they are super clear on what they do

code/index.html
```diff
@@ -0,0 +1,427 @@
...
+        </section>
+
+        <section>
+            <!-- ***featured projects ***  -->
+            <div class="headings-container">
+                <h2 class="section-heading featured">featured projects</h2>
+            </div>
+            <div class="auto-colums" tabindex="0">
+                <article class="content-container">
+                    <a href="https://marysnopok-todos.netlify.app" tabindex="0" role="button" aria-pressed="false"
```
madeleinesvensson: also very nice that you added the tab index and the role! I used a screen reader and I could navigate everywhere and the screen read everything!

code/script.js
```diff
@@ -0,0 +1,5 @@
+const themeToggle = document.getElementById("theme-toggle");
```
madeleinesvensson, line 1: love this! it even worked with the screen reader to toggle

madeleinesvensson (commented):
This portfolio looks soooo good Mariia! I love the header that changes with the dots and the fact that you can toggle colors! very nice feature! I used voice over and navigated and managed to go everywhere and it was very nice! I also found out in your code that you made changes on hover which I hadn't seen when I looked at it and that made it very interactive and more alive so good job! The only thing i would suggest is to remove cometed out code in the css if you don't use them to make it more clear! Otherwise good job with the code(clean and readable) and the naming of classes (also intuitive and reusable) and be very proud of this project! it looks very good!

code/canvas.js
MarySnopok: Thank you ) it was cool to play with it

code/index.html
MarySnopok: It was less complicated as I expected just had an issue with picture resolution and path to it

code/index.html
MarySnopok: yep it is the div with dots and the code for it in the canvas.js file

code/index.html
MarySnopok: lol ))) thats means like less than 6 years of experience . I took this approach from the figma design , but yep it is confusing I might take it away .

code/index.html
MarySnopok: Thanks thats my key to dont get made while styling lol

code/script.js
MarySnopok: Really )? I was not even expecting it lol ) Great to know ! Thank you for checking it out !

PR #71, opened 2021-10-09
Portfolio

ghost:


MariaThomasson (changes requested):
I started the code review by testing your site in two different validators and my suggestion is that you have a look at them and correct some of the errors before I continue with the review. If you don´t have time or want to do it just let me know and I will continue the review. Im pasting in the links from the twoI tested, let me know if there is any problem and I can help you :)

[Jessica Review week 8.docx](https://github.com/Technigo/portfolio/files/7322181/Jessica.Review.week.8.docx)

PR #81, opened 2021-10-10
Maria Petersson Portfolio

hejmaria:
The assignment was to create a responsive website, mobile-first, and follow a given design. The key focus was to make sure the site is accessible. 

It was a big assignment, taking many hours to write. I focused on doing one section at a time, making sure it was well done before going on the next. After a while I needed to work broader, working more with the common style or code for the sections. I didn't make any great mistakes, but in the end, I needed to make some changes in the structure to make it follow the design for screens. It was a challenge. 

Due to life and illness, I was short of time and needed to prioritize. Would love to work further with the content. 

Deployed site at:
https://mariapeterssonportfolio.netlify.app/


Groovedharry (approved):
You have very easy code to read and it all makes sense. You have achieved all blue requirements.

Your code is very clean and I will look through it more as I believe I can learn abit from looking at it. Good job!

PR #83, opened 2021-10-10
Julia N portfolio

JuliaNiki:
https://julia-nikitina-portfolio.netlify.app


index.html
```diff
@@ -0,0 +1,261 @@
...
+        <section class="header-text">
+            <h1 class="title-my-name">
+                <span class="portfolio">PORTFOLIO:</span>
+                JULIA NIKITINA
+            </h1>
+            <h1 class="title-my-background">frontend developer</h1>
+            <h1 class="title-plus">+ logistics & project management background</h1>
+        </section>
+        <img src="assets/me.jpeg" class="profile-pic" alt="Image of Julia" />
+        <nav class="social-medias">
```
elsisco, line 32: Nice that you used the semantic tag for navigation!

index.html
```diff
@@ -0,0 +1,261 @@
...
+            </a>
+        </nav>
+    </main>
+    <section class="about-me">
+        <p class="text-about-me">
+            <span class="line"></span>
+            <span class="dot"></span>
+            Hi! I am Julia, a frontend developer with 5 years experience within supply chain and project
+            management. I am
+            a detail-oriented person with a structured and creative approach for problem-solving and great comminication
```
elsisco, line 51: Small spelling mistake here on communication. Wouldn't comment on it normally but since it is for companies to read :) I'm using the Code Spell Checker extension in VS Code and I think it works great!

index.html
```diff
@@ -0,0 +1,261 @@
...
+            </ul>
+        </div>
+    </section>
+    <section class="for-more">
+        <h2 class="title-section">FOR MORE</h2>
+        <div class="social-media-container">
+            <div class="link-social-media">
+                <a href="https://www.linkedin.com/in/julia-nikitinashlm/" class="social-media-button-big"
+                    target="_blank">
+                    <i class="fa fa-linkedin"></i>
```
elsisco, line 386: Hmm, what is the purpose of the <i> in this section? I thought they were for producing italic text but there is no text here so i guess not :)

index.html
```diff
@@ -0,0 +1,261 @@
...
+                </a>
+                <p class="social-media-text">GITHUB</p>
+            </div>
+
+        </div>
+    </section>
+    <footer class="footer">
+        <h3 class="contact-title"> CONTACT</h3>
+        <h4 class="contact-info">Julia Nikitina</h4>
+        <a href="tel:+46723417366" class="contact-info">+ 46 72 341 73 66</a>
```
elsisco, line 409: Great that you made the phone number and email to clickable links!

style.css
```diff
@@ -0,0 +1,513 @@
...
+    font-family: 'Montserrat', sans-serif;
+    font-weight: 400;
+    font-size: 17px;
+    line-height: 29px;
+    text-align: center;
+}
+/* styling for all containers to have as flexbox and display as column */
+.projects-container,
+.my-thoughts,
+.other-projects {
```
elsisco, line 147: Great that you are grouping elements with similar styling - so much easier to adjust if necessary!

style.css
```diff
@@ -0,0 +1,513 @@
...
+.image-container {
+    width:100%;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    position: relative;
+}
+.project-image {
+    width: 100%;
+    height: 250px;
```
elsisco, line 186: I think this measurement is what makes the thoughts-image distort.

elsisco (commented):
Review-time! :D Thank you for letting me having a look at your project. 
You have followed the design files provided really well and the code is well written and easy to understand. Good job! I have a few suggestions for some minor things I noticed.
* The design spec doesn’t say what to do about tablet so it is up to us if we want to display it somewhat like the phone (column) or like the desktop (row wrap). The tablet is about twice as wide a phone and I would say it wouldn’t be a bad idea to display it like the desktop, or at least like desktop when the tablet is throttled (90°). If you’d like you could start the desktop display already at 1024px screen width (throttled iPad), or even 768px, but it is down to taste really. 
* The text under TECH has no width specified so the lines of text get really long on wide screens. To improve readability, maybe keep it at the same width as the about me text?
* The image under thoughts displayed on mobile view gets distorted because only the width is flexing with the screen size. 
Other than that, nothing else to say. Your portfolio simply looks awesome! Great job Julia!

index.html
JuliaNiki: oh, thank you for this tip! 👍🏼

PR #91, opened 2021-10-10
Madelene Trang's portfolio

MT-dotse:


jessicatf (commented):
Hi Madelene,

Very nice job indeed! Looks great. You should be proud. You have met the requriements!
The only tiny thing I could see is that the images (in featured-projects) in a larger screen are not "even" and are not exactly the same size.

Best regards,

Jessica Falk

PR #92, opened 2021-10-10
Week 8 | Portfolio | Amanda Tilly

amandatilly:
# portfolio

Creating my portfolio with HTML and CSS.

## The problem

Focused on the blue requirements this week. I was unwell most of the week and in the end, I only had two days to finish the project, there is much still to be done but I am happy with what I have for now - the bones to keep building on.

## View it live

https://amandatilly.netlify.app/


code/index.html
```diff
@@ -0,0 +1,405 @@
...
+            target="_blank"
+          >
+            <img
+              class="stack-icon"
+              src="./assets/stack-icon.png"
+              alt="stackoverflow icon"
+            />
+          </a>
+        </nav>
+        <div class="hero-text">
```
Fatima-GR, line 73: Just to make it more clear and be consistent,  I would have named this div, hero-text-container, since below you have another div that is called profile-image-container.

code/index.html
```diff
@@ -0,0 +1,405 @@
...
+            />
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse
+            ornare quis purus ut pulvinar. Sed semper convallis tincidunt. Etiam
+            porttitor.
+          </p>
+        </section>
+      </section>
+    </header>
+
+    <main>
```
Fatima-GR, line 125: very nice using the main tag, it's great for accessibility.

code/index.html
```diff
@@ -0,0 +1,405 @@
...
+
+      <section class="portfolio-section">
+        <div class="">
+          <div class="portfolio-heading-container">
+            <h1 class="portfolio-heading">skills</h1>
+          </div>
+          <section class="skills">
+            <div class="skills-container">
+              <div class="skills-card">
+                <ul>
```
Fatima-GR, line 504: Well done in using li tags instead of span tags.

code/index.html
```diff
@@ -0,0 +1,405 @@
...
+        </div>
+      </section>
+
+      <section class="portfolio-section">
+        <div class="container">
+          <div class="portfolio-heading-container">
+            <h1 class="portfolio-heading">for more</h1>
+          </div>
+          <section class="more">
+            <div class="more-icons">
```
Fatima-GR, line 560: I would have used a nav tag since it's a navigation section.

code/index.html
```diff
@@ -0,0 +1,405 @@
...
+          </section>
+        </div>
+      </section>
+
+      <section class="contact">
+        <div class="container">
+          <div class="portfolio-heading-container">
+            <h4 class="contact-heading">contact</h4>
+          </div>
+          <section class="contact-text">
```
Fatima-GR, line 601: It would be great if you could make this section clickable as well, so this way, recruiters would have it easier to contact you. For example, I would suggest making your name clickable and leading to your  Linkedin. And your email is the same, when the recruiter clicks on your email to send you a message, it would be great if it would lead you directly to the email page. I did that on my Github project, so you can check it out but here is how I did it: <a href="mailto:insertemailhere@xyz.com">Email</a>. Just type, after the malito word, your email, and then, between the anchor tags, your email address one more time.

code/style.css
```diff
@@ -0,0 +1,637 @@
+body {
+  margin: 0;
```
Fatima-GR, line 2: In the body tag or in the universal selector (*) remember always to add 3 properties, margin:0, padding:0, box-sizing: border-box.

code/style.css
```diff
@@ -0,0 +1,637 @@
...
+  font-weight: 700;
+  font-size: 17px;
+  line-height: 29px;
+  text-align: center;
+  color: #fa382f;
+  width: auto;
+  text-transform: uppercase;
+}
+
+ul {
```
Fatima-GR, line 489: Maybe use classes instead of using HTML tags, as they have less power than clases.

Fatima-GR (commented):
Good job Amanda! You made this project in such a short period of time and that's something to admired! 


In terms of responsiveness, your project is responsive, except on 320 px, the page does not fill all the width of the screen so check that out! Also, remember that the featured section only should show 2 projects on the phone approach and 4 in the desktop approach. You can get that easily with display none and then display: flex in the media queries. Remember also to make all the featured sections clickable. You can get that by wrapping all the cards with an anchor tag. <a> . 


In general, your project looks really nice, you chose a nice image that describes your personality very well. Your project is accessible as you used HTML semantic tags and it is very well organized. Congrats Amanda 😊

PR #101, opened 2021-10-11
Week 8, Portfolio, Maria Thomasson

MariaThomasson:
https://mariathomasson-portfolio.netlify.app/


code/index.html
```diff
@@ -0,0 +1,263 @@
...
+    </section>
+
+    <section class="contact-section" tabindex="0">
+        <h4 class="contact-header">CONTACT</h4>
+        <p>Maria Thomasson</p>
+        <p>mariathomasson@hotmail.se</p>
+    </section>
+
+
+
```
ghost, line 282: Really good work Maria! Love your site. The hover effect and small details like the github icon link under the projects are amazing! Nothing looks or feel "broken", the only I think you can add is traget="_blank" to the link-tags so a new page are opening when you click on a link (but that are a preferens thing so dont mind if you dont like that function)!
The code is easy to follow as well!
The performance score is 95% in ligth house(chrome), really good! On https://wave.webaim.org/ I dont get any error just some errors on contrast - but as I know you allready know this and I think you shuld keep the colors/design as it is! Nice flow on the tab-function as well!  https://validator.w3.org/ - no errors!! Good work!!!
- [ YES ]  Am I able to understand the code easily?
- [ NO ]  Does the code have lots of duplicates? Could it be tidied up?
- [ YES ]  Does it work or does anything look broken?
- [ YES ]  Does it follow the general and blue requirements?

PR #113, opened 2021-10-12
My portfolio

Alisebrink:


styles.css
```diff
@@ -0,0 +1,434 @@
...
+h3,
+h4,
+h5,
+h6 {
+  margin: 0;
+}
+
+p,
+li,
+a {
```
MT-dotse, line 21: That is really clever to structure the fonts and avoid repeating code.

styles.css
```diff
@@ -0,0 +1,434 @@
...
+.text__container {
+  position: absolute;
+  top: 121px;
+  line-height: 18px;
+  width: 250px;
+}
+
+.text__container h1 {
+  line-height: 1.2em;
+  font-size: 3em;
```
MT-dotse, line 105: It is good to be consistent with using either em or px.

index.html
```diff
@@ -0,0 +1,310 @@
...
+                    A chatbot made with pair-coding.</span
+                  >
+                  >>
+                </h4></a
+              >
+              <span class="skill">JavaScript ES6</span
+              ><span class="skill">CSS3</span>
+            </div>
+            <div class="other-project">
+              <a target="_blank" target="_blank"
```
MT-dotse, line 245: there is a duplicated  target="blank"

MT-dotse (commented):
Amazing work Linnea, 

I can see that you are very skilled with both HTML and CSS, in both the structure and the styling. Both HTML and CSS code looks clear with no repetition and it is easy to follow. 

As for the finished website, you have picked a nice color for the headers that suits well with your hero image. I also like how well structured all the sections are. 

The HTML code looks clear with no repetition and the CSS is easy to follow. 

The site is responsive and works on various displays.

The website has been tested in https://validator.w3.org/ and https://jigsaw.w3.org/css-validator/  and only one minor error has been found. A duplicated “target_blank” on line 171 in the HTML-code. 

Regarding the requirements you have met all the blue.
Well done! You should be proud!

PR #130, opened 2022-03-06
portfolio-åsa-sieurin

sieurin:
(FYI: I pitched my own design for technigo, that's why my layout looks slightly different).

I'm not done with everything I want (“color mode” in the navigation is in the making, so the buttons are still empty but will have a toggle function later). Same goes with some content etc.

But I think this portfolio will be an ongoing  project, and at the end of the day I am actually pretty happy with how far I have come these days..


jjwwkkmm (commented):
I am impressed by your portfolio design and the courage to commit to a new idea and complete it. And by complete I mean making it work well in the code as well. The code is really structured and is making use of HTML5 tags, and I can see the styling is making use of cross-browser compatible tags. I also see you are making use of complex CSS stylings like animation, var() - which I haven't even come across myself.

Don't really have much to say, I think both your code and design is great.

PR #137, opened 2022-03-06
Week #8 : Portfolio

rawisou:
# portfolio

I need to work on hover effects and more when I have more time! 

View it live: https://rawi-portfolio.netlify.app/


bostrombundock:
Rawi, it is very easy to follow your code, even if it not any comments in the code! I might be totally wrong hence I have not grasped the responsiveness fully, but when I look in the inspector tool in mobile view I cant see your whole "hello text", and further down by the more thougts, its a bit tight, so maybe it need some margin or just an adjustment in the @media.

PR #143, opened 2022-03-06
Sofia Ringstedt portfolio

sofiaringstedt:
https://sofiaringstedtportfolio.netlify.app/


style.css
```diff
@@ -0,0 +1,413 @@
...
+}
+
+.profile-image {
+  width: 180px;
+  border: 2px solid #FFF; 
+  border-radius: 50%;
+}
+
+/* styling of span, creating the arrow in summary section*/
+.arrow {
```
EmmaSprings, line 171: For this icon, it is possible to download an svg or png for figma in this case, makes it easier as it can then be added next to the <p> and then styled in css with margins/padding/width etc whats needed :) But impressed with this to say the least!

style.css
```diff
@@ -0,0 +1,413 @@
...
+.tag {
+  background: #FDDFE7;
+  font-family: 'Roboto';
+  font-weight: 400;
+  font-size: 9px;
+  margin: 0px 5px 42px 0px;
+  padding: 3px 6px;
+}
+
+.project-container h4, .thoughts-container h4 {
```
EmmaSprings, line 274: This is clever, need to learn from this!

EmmaSprings (approved):
Well done with your portfolio! The design has been followed from Figma very well and the responsiveness works as expected. 
One small thing I noticed was that the overlay header on the project image wasn't centered, if you don't want that, I tried to figure out a few changes in the code that seemed to work :) Same thing for the "for more" icons. 
 So, for the header "chat bot" and "github tracker", you could maybe try this if you want to: 

.project-title {
	.	     width: 100%;  // NEW
	.	     position: absolute;
	.	     top: 35%;  // NEW
	.	     //*left: 15%//* REMOVE
	.	     font-size: 25px; //This can still be 28px 
	.	     text-align: center; //NEW
	.	     letter-spacing: 0.01em;
	.	     color: white;
	.	     text-transform: uppercase;

And for the icons in "For more" section I tried this

.more-links-tab {
	.	     text-decoration: none; 
	.	     font-size: 11px;
	.	     font-weight: bold;
	.	     line-height: 13px;
	.	    *// text-align: center;*// REMOVE
	.	     color: #F14A16;
	.	     text-transform: uppercase;
	.	     width: 100px;  //NEW
	.	     display: flex;  
                    * //margin: 10%; *//REMOVE
	.	     flex-direction: column;
	.	     align-items: center;
}

This might not be perfect, but it seemed to work in the console, so thought I'd write it down here as a tip!

Exciting to see what will be added next in the portfolios!

sofiaringstedt:
Thank you so much @EmmaSprings for your very helpful comments!! They def helped, especially the ones regarding centering. Have a nice weekend!

PR #153, opened 2022-03-06
Project 8 - Portfolio

mathildakarlsson:


code/index.html
```diff
@@ -0,0 +1,210 @@
...
+    <link rel="stylesheet" href="style.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link href="https://fonts.googleapis.com/css2?family=Montserrat:wght@100;200;300;400;500&family=Roboto:wght@100;300;400;500;900&display=swap" rel="stylesheet">
+    <title>Mathilda Karlsson, Developer Portfolio</title>
+</head>
+<body>
+    <header class="header">
+        <div class="container header-container">
+            <div class="header-image-group">
```
SteppbySteph, line 17: In the header-image-group I see you used a div with two elements (an <a> tag with image of three icons and then the actual profile picture). When you have added the different images/icons for the social media you could wrap these three in one div and maybe let the profile-img act as a separate element. In this way it might be easier to style and place them according to your design of choice. This layout you have done so far is looking splendid!

code/index.html
```diff
@@ -0,0 +1,210 @@
...
+            </div>
+            <div class="header-text-group">
+                <h1 class="header-name"><span class="header-portfolio">PORTFOLIO</span> Mathilda Karlsson</h1>
+                <h1 class="header-title">frontend developer</h1>
+            </div>
+        </div>
+    </header>
+    <section class="section">
+        <div class="intro-section container">
+            <article class="portfolio-description">
```
SteppbySteph, line 52: In this section you are using an article-tag and div-tag before you add a paragraph to the section. To write this part with less code you could exclude the above named tags. But I am wondering if you did this for a reason. If so don't mind this comment :)

code/index.html
```diff
@@ -0,0 +1,210 @@
...
+                <div class="portfolio-description-group">
+                    <p> Hello! I’m a strategic and creative senior art director/digital designer with frontend developer skills and experience design at. I’m all about solving problems that matter and creating awesome, meaningful products for people to love.</p>
+                </div>
+            </article>
+        </div>
+    </section>
+    <section class="section">
```
SteppbySteph, line 60: I see now that you are using similar structure for each section. This can mean that there are more styling to add in the CSS. But I will have this in mind when I keep reviewing your code. Looking at the mobile version you have delivered a very good result!

code/index.html
```diff
@@ -0,0 +1,210 @@
...
+        <div class="for-more-section container">
+            <article class="for-more">
+                <div class="for-more-description-group">
+                    <h2 class="for-more-header">FOR MORE</h2>
+                    <div class="for-more-images">
```
SteppbySteph: An alternative is to make a "div-tag"  and set the "a-tag" and the "p-tag" as children element. In this way they appear as one element in the container. If you repeat this for Github and StackOverflow you get three "div" in total. Set the display: flex; justify:content to the "for-more-description-group".  Hopefully this will give you a result you can use. You are almost done! Way to go!

code/index.html
```diff
@@ -0,0 +1,210 @@
...
+                        <p class="github-title">GITHUB</p>
+                        <p class="stack-title">STACK OVERFLOW</p>
+                    </div>
+                </div>
+            </article>
+        </div>
+    </section>
+    <section class="section">
+        <footer>
+            <div class="footer-container"></div>
```
SteppbySteph: Make a parent div after the footer-container and add the "contact" as a separate heading-element and another div as a child element where you contain three paragraphs with the info. Style this in the new div-container (not the footer-container)  as: 
display:grid;
grid-template-columns: 1 fr;
 justify-content: center;
 text-align: center;
 align-items: center;
  color: white;

Hopefully you understood this and that it will help you out!

SteppbySteph:
You have a beautiful color palette to your portfolio and the content is almost completed. I have added some tips on how to work with the "for more" section and "footer". You are almost done so hang in there! This was not an easy task but you came a long way in a few days before the demo. You can do this, Mathilda! Your portfolio is gorgeous!

PR #155, opened 2022-03-06
Project portfolio Emma Springs

EmmaSprings:


code/style.css
```diff
@@ -0,0 +1,489 @@
...
+.section:nth-child(2n) {
+    background-color: rgba(252, 229, 190, 0.173);
+}
+
+ header {
+     background-image: url(./images/cyan-2.jpg);
+     filter: saturate(180%);
+     background-size: cover;
+     background-position: top;
+	 height: 520px;
```
sofiaringstedt, line 21: Are there a reason for these intendentions? Just curious :)

code/style.css
```diff
@@ -0,0 +1,489 @@
...
+	 position: relative;
+ }
+
+
+.container {
+    display: flex;
+    flex-direction: column;
```
sofiaringstedt, line 29: Here I am thinking you could add justify-content: center; to fix the desktop responsiveness (especially at More Thoughts

code/style.css
```diff
@@ -0,0 +1,489 @@
...
+
+    .author-image-container {
+        left: 45%;
+    }
+    
+}
+
+@media screen and (min-width: 1200px) {
+
+    .section:nth-child(2n) {
```
sofiaringstedt, line 469: How does this work, that the section colouring is in the medie query?

code/index.html
```diff
@@ -0,0 +1,280 @@
...
+                            <p tabindex="0">node.js</p>
+                        </div>
+                    </div>
+                    <div class="skills">
+                        <h2 tabindex="0" class="skills-title">toolbox</h2>
+                        <div class="skills-text">
+                            <p tabindex="0">Adobe Photoshop</p>
+                            <p tabindex="0">Adobe InDesign</p>
+                            <p tabindex="0">Lightroom</p>
+                            <p tabindex="0">Figma</p>
```
sofiaringstedt, line 379: Ah I didnt know you could use tabIndex for other elements than links!

code/images/start-figure.svg
```diff
@@ -0,0 +1,9 @@
+<svg width="55" height="11" viewBox="0 0 55 11" fill="none" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
sofiaringstedt, line 1: would love to know how this works!!

sofiaringstedt (commented):
Really nice looking website that functions well! I am not 100% sure how strict Technigo will be on the requirement that you follow the Figma design, therefore I would recommend you to take a look at your More thoughts section. 

A general tip would be to use the fact that you have a nice section/container structure to your advantage and put some joint styling on these as much as you can. 
It also looks like you could clean up some of your unused files (mostly images) and I would have loved some more comments in your CSS code to better follow and learn from you! 

Other than that - great job - its a really nice looking website! However, I would look at your desktop styling asap, it’s just a small fix that is probably quickly done! It was nice looking at your code and your website.

code/style.css
EmmaSprings: I'll try that :) However, when I look at the desktop view, everything looks fine. Could it be that you see the tablet view, as it is up to 1200px? I need to fix that one better for sure. Work in progress :)

code/style.css
EmmaSprings: In the Figma file, the desktop had a different opacity for the background color than in the mobile view. Maybe not that important though :)

code/style.css
EmmaSprings: That would just be a "typo" haha. will fix.

code/images/start-figure.svg
EmmaSprings: Not sure I understand it right how you mean, but I think this is just code for the image, it's there by default so nothing I've done myself :)

PR #165, opened 2022-03-06
Project week 8 – Portfolio

emmajosefina:
https://emmalindell.netlify.app/


JenFi72:
Very good work with this project!! Love the design and looks very good in all responsive moods. 

The code is very structured ad easy to follow. Maybe just add some comments on some places. Emma reaches blue level requirements by far building the page with HTML and CSS. The page is built in a cohesive way using flexbox and "re-using" sections and divs in order to keep code as possible. Great job!

PR #166, opened 2022-03-06
Project portfolio-week 8

EmmaaBerg:


index.html
```diff
@@ -0,0 +1,283 @@
...
+        rel="stylesheet">
+    <title>Emma Berg's Portfolio</title>
+</head>
+
+<body>
+    <header class="header">
+        <div class=" container header-container">
+            <div class="header-links">
+                <!--Need to fix and add my linkedin here-->
+                <a arial-label="Go to Linkedin profile" href="#" target="_blank" title="LinkedIn Emma Berg">
```
nadialefebvre: Good use of aria-label here.

index.html
```diff
@@ -0,0 +1,283 @@
...
+                </div>
+                <div class="subtitle-box">
+                    <h2 class="subtitle">Other projects</h2>
+                </div>
+                <div class="other-projects">
+                    <!--ADD LINK TO GITHUB REPO-->
+                    <div class="other-project-box">
+                        <h3 class="other-projects-title">ADD TITLE OF PROJECT</h3>
+                        <p class="other-projects-description">ADD DESCRIPTION HERE
+                            <span class="arrows">&gt;&gt</span>
```
nadialefebvre: Here a tip: you could add `aria-hidden="true"`. The result will be that screen readers won't read the arrows. Now, it reads "greater than greater than" and people who can't see don't need to hear that.

index.html
```diff
@@ -0,0 +1,283 @@
...
+                </article>
+            </div>
+        </div>
+    </section>
+
+    <!--FOR MORE SECTION-->
+    <section class="section">
+        <div class="container">
+            <div class="title-box">
+                <h2 class="title">FOR MORE</h2>
```
nadialefebvre, line 462: In this section, the icons aren't centered under the heading, so maybe you could look into your CSS to make it look better.

index.html
```diff
@@ -0,0 +1,283 @@
...
+                    </a>
+                    <div class="tag-box"><span class="tag">HTML</span><span class="tag">CSS</span>
+                        <span class="tag">JavaScript ES6</span>
+                    </div>
+                </div>
+
+            </div>
+    </section>
+
+    <!--MY THOUGHTS ABOUT CODE-->
```
nadialefebvre, line 381: Great that you made this structure for when you will have something to put in.

style.css
```diff
@@ -0,0 +1,484 @@
+
+/*Add global varibles for color here */
+:root {
+    ---primary: #CD8987;
```
nadialefebvre: Great with the colours variables, it makes it easier to change the colours, smaller risk of mistake.

style.css
```diff
@@ -0,0 +1,484 @@
...
+
+h1 {
+    font-size: 40px;
+    font-weight: 700;
+    line-height: 45px;
+    margin: 0;
+}
+
+h2 {
+    display: table;
```
nadialefebvre, line 31: I wonder what this `display: table` does? I never used that (yet).

style.css
```diff
@@ -0,0 +1,484 @@
...
+
+
+/*------ MEDIA QUERIES ------*/
+@media screen and (min-width: 768px) {
+
+h1 {
+    font-size: 60px;
+    line-height: 80px;
+}
+.container {
```
nadialefebvre: Indentation lost me a bit here, I thought we weren't in the media queries anymore ;)

style.css
```diff
@@ -0,0 +1,484 @@
...
+.project-overlay-title {
+    color: var(---main);
+    background-color: transparent;
+    font-size: 28px;
+    position: absolute;
+    top: 45%;
+    left: 0;
+    bottom: 0;
+    right: 0;
+    transition: 0.5s ease;
```
nadialefebvre, line 241: Nice with the transition.

nadialefebvre (commented):
You did a great job, Emma! When I look at your portfolio, I find it relaxing, maybe it's the colours :) You seem to have followed quite well the requirements, so be proud of your work!

index.html
EmmaaBerg: Thank you, I will definitely try to add that :D

index.html
EmmaaBerg: Yes I struggled with those icons but will take a look at it today

style.css
EmmaaBerg: As I understand it creates/defines a block-level box, just like when you write `display: block` and behaves like the `<table>` element in HTML

style.css
EmmaaBerg: Oops, will look at the indentation!

PR #170, opened 2022-03-07
Week 8- Portfolio-Sara Boström-Bundock 

bostrombundock:


index.html
```diff
@@ -0,0 +1,293 @@
...
+                                    <h2 class="project-image-text">
+                                        NEWS SITE
+                                    </h2>                     
+                                </div>                                                    
+                            </div>
+                        </div>
+                        <h4>NEWS SITE built in HTML and CSS</h4>
+                        <p>Responsive web site using HTML and CSS.</p>
+                    </a>
+                    <ul class="tags">
```
rawisou, line 136: I like your thinking here putting the list for tags. :)

style.css
```diff
@@ -0,0 +1,432 @@
...
+    width: 49%;
+    text-align: left;
+}
+
+.project-item {
+    position: relative;
+    margin-bottom: 20px;
+}
+
+.project-item::before {
```
rawisou, line 223: Pseudo classes fit in very well in this case! Very nice!

rawisou (commented):
Hello Sara,

Your website works as it should. You've followed the design and fulfilled the project's requirements.

Your code is clean and well formatted. You have named classes in a semantic way which makes it easy for others to understand.

If there's one thing that would make your portfolio looks even closer to the design is that you could offset the default setting for the margin in` body.`

You could add something like this in your existing `body` (line 1 in style.css).

```
body {
margin: 0;
}
```

In your case,  this would be: 
```
body {
    margin: 0;
    font-size: 18px;
    line-height: 30px;
    font-family: 'Montserrat', sans-serif;
    font-style: normal;
    font-weight: 400;
}
```

PR #172, opened 2022-03-07
project-portfolio-JaEngd

JaEngd:


Nosslexa (commented):
Great job whit your Portfolio!
Love to see a portfolio whit its own style, really like the rolling text! Your site works as it should and your code looks clean is easy to follow. You fulfill the requirements of this project, well done!

PR #175, opened 2022-03-08
Portfolio - Tiina Liukkonen

tiiliu:


index.html
```diff
@@ -0,0 +1,380 @@
...
+     <header class='header'>
+         <div class='header-container'>
+
+             <div class='header-icon-wrapper'>
+                <img class='header-icon-image' src='./images/linkedin-icon.svg' alt='linkedin icon'>
+                <img class='header-icon-image' src='./images/github-icon.svg' alt='github icon'>
+                <img class='header-icon-image' src='./images/stackof-icon.svg' alt='stack overflow icon'>
+             </div>
+
+            <div class='header-image-group'>
```
emmahogberg88, line 34: A bit too detailed perhaps.. but you can add one more space here for the indentation :)

index.html
```diff
@@ -0,0 +1,380 @@
...
+                <h1 class='header-subtitle'>with creative ambition</h1>
+             </div>
+
+         </div>
+    </header>
+
+<main>
+
+ <!--INTRO AND TECH SECTION-->   
+    <section aria-label='' class='section'>
```
emmahogberg88: aria-label is empty, maybe add a comment so you don't forget to add content :)

emmahogberg88: Just curious.. Is there a certain reason you add a class here? Is it for specificity? 🤔

index.html
```diff
@@ -0,0 +1,380 @@
...
+        <div class='container'>
+
+                <p class='intro-text'>
+                    <span class='intro-line'></span><span class='intro-dot'></span>
```
emmahogberg88: Love this section, the text about you is on spot!

index.html
```diff
@@ -0,0 +1,380 @@
...
+                            <p>API</p>
+                        </div>
+                        </a>
+                    </div>
+                </div>
+
+                <div class='project-card'>
+
+                    <div class='project-image-group'>
+                        <a class='nav-item' href='https://github.com/tiiliu/project-weather-app' target='-blank'>
```
emmahogberg88: target='_blank' instead of target='-blank'?

emmahogberg88: Same on row 110 and many other rows. Tried to google so maybe it´s correct but I could just found that it is target='_blank'

index.html
```diff
@@ -0,0 +1,380 @@
...
+                    to bring those three together to create visually appealing and user-friendly products. 
+                </p>
+
+        </div>
+    </section>
+
+    <section class='section'>
+        <div class='container'>
+
+                <h2 class='tech-section-title'>TECH</h2>
```
emmahogberg88: A Kevin Powell tips I heard of is to add a class that only make the text uppercase, instead of writing the text itself i caps. :)

index.html
```diff
@@ -0,0 +1,380 @@
...
+                        <a class='nav-item' href='https://github.com/tiiliu/project-weather-app' target='-blank'>
+                        <p class='project-title-overlay'>WEATHER APP</p>
+                        <img class ='project-image' src='./images/weather-app.png' alt='image displaying a preview of the weather forecast web site' />
+                        <div class='project-image-overlay'></div>
+                        </a>
+                    </div>
+                    
+                    <div class='project-text-group'>
+                        <a class='nav-item' href='https://github.com/tiiliu/project-weather-app' target='-blank'><span class='link-styler'></span>
+                        <h4 class='projects-small-title'>WEATHER APP CREATED WITH WEATHER APIS</h4>
```
emmahogberg88: I like the class name "link-styler" I will use that too.

index.html
```diff
@@ -0,0 +1,380 @@
...
+                </div>
+
+            </div>
+       
+            <h3 class='more-thoughts-title'>MORE THOUGHTS</h3>
+            <div class='more-thoughts-container'>
+                <div class='more-thoughts-group'>
+                    <span class='dates'>SEP 2019.</span>
+                    <span class='blog-small-title'>5 Apps that will make coding easier.</span>
+                    <span class='arrow'> >> </span>
```
emmahogberg88: Maybe use semantic HTML here, like a paragraph tag to wrap around our span-tags.

index.html
```diff
@@ -0,0 +1,380 @@
...
+<!--SKILLS SECTION-->
+    <section class='section'>
+        <div class='container'>
+
+            <h2 class='skills-section-title'>SKILLS</h2>
+
+            <div class='skills-parent'>
+
+                <div class='skills-child'>
+                    <ul class='small-text-skills'>
```
emmahogberg88, line 710: I like that you use a list for this! 👍

style.css
```diff
@@ -0,0 +1,623 @@
+* {
+    margin: 0;
+    padding: 0;
+}
+
+:root {
```
emmahogberg88, line 6: This is really smart!

style.css
```diff
@@ -0,0 +1,623 @@
...
+    align-items: center;
+    justify-content: center;
+    text-align: center;
+}
+
+.arrow {
+    letter-spacing: 0.01em;
+}
+
+.tag-wrapper p {
```
emmahogberg88, line 151: Maybe you can add a flex-wrap:wrap to make the tags wrap, otherwise they stretch vertically to fit :) (look at chat-bot on 375px to see what I mean).

style.css
```diff
@@ -0,0 +1,623 @@
...
+.featured-projects-container, .other-projects-container {
+    width: 100%;
+}
+
+
+.project-card-margin {
+    margin-bottom: 32px;
+}
+
+.project-image-group {
```
emmahogberg88, line 304: Your images stretches out on desktop, so maybe add a max-width here :)

style.css
```diff
@@ -0,0 +1,623 @@
...
+}
+
+.hidden-from-mobile {
+    display: none;
+}
+
+
+/****************TEXT STYLING***************/
+
+p, span, h4, .header-author-title, .dates, .blog-arrow, ul, .icon-text  {
```
emmahogberg88: Another Kevin Powell tips is to add each element on a new row and not in one line, it makes it a bit more easy to read and find in the css file :)

index.html
```diff
@@ -0,0 +1,380 @@
...
+                        <p>HTML</p>
+                        <p>CSS3</p>
+                        <p>JavaScript ES6</p>
+                        <p>Pair-programming</p>
+                    </div>
+                    </a>
+                </div>
+
+                <div class='other-projects'>
+                    <h4 class='projects-small-title projects-title-desktop'>PROJECT TITLE</h4>
```
emmahogberg88: In the previous "other project" you have an anchor-tag. This section has a different hover-effect than the previous "other-object".

emmahogberg88 (commented):
Really good job Tiina! 🎆 Your page looks super good and all the details are there! I have done my best to find things that might help you improve your project (it wasn't easy.. 😄) There are some Kevin Powell tips (but I guess there are more ways to do it the "right" way, so go for what suits you best). 
I think you did a really good job and the code is very well structured and there´s a logic order.

index.html
```diff
@@ -0,0 +1,371 @@
...
+                        <p class='icon-text'>GITHUB</p>
+                    </a>
+                </div>
+
+                <div class='for-more-icon-container' tabindex='0'>
+                    <a class='nav-icon' aria-label='Visit the authors stack overflow profile' href='https://stackoverflow.com/c/technigo/users/310' target='-blank'><span class='link-styler'></span>
+                        <img alt='stack overflow icon' class='for-more-icon-image' src='./images/stackof-icon.svg' aria-hidden='true'>
+                        <p class='icon-text'>STACK OVERFLOW</p>
+                    </a>
+                </div>
```
emmahogberg88: I noticed that this part is not centered since the stackoverflow icon is taking more width than the other icons. Maybe you can add a set width to each "for-more-icon-container" so the space between has same width :)

index.html
```diff
@@ -0,0 +1,371 @@
...
+
+                    <div class='blog-text-group'>
+                        <p class='dates'>DEC 2019</p>
+                        <h3 class='blog-title'>How to stop being scared: Learning to love code</h3>
+                        <p class='blog-text'>
+                            Writing code is less about “just working”. Instead, it's much more about
+                            creativity. It's similar to creative writing in that the author architecting
+                            a story must use an assortment of tools and components (i.e., content) in a
+                            way to satisfy a particular objective.<span class='blog-arrow'> >> </span>
+                        </p>
```
emmahogberg88: This part misses tabindex :)

index.html
```diff
@@ -0,0 +1,371 @@
...
+
+                <div class='project-card project-card-margin' tabindex='0'>
+                    <a class='nav-item' aria-label='Visit the github repository of project github tracker' href='https://github.com/tiiliu/project-github-tracker' target='-blank'><span class='link-styler'></span>
+                        <div class='project-image-group'>
+                            <p class='project-title-overlay'>GITHUB TRACKER</p>
+                            <img alt='image displaying a screenshot of the movie chatbot app' class ='project-image' src='./images/gitfolio.png' tabindex='0' />
+                            <div aria-hidden='true' class='project-image-overlay'></div>
+                        </div>
+
+                        <div class='project-text-group' tabindex='0'>
```
emmahogberg88: For some reason, when I tab through your site using Chromes screen reader, the projects content repeats itself. So it says the same thing twice.. Maybe it's because of the tabindex='0' on both line 73 and line 74 but I'm not sure. It's the same for all projects that it repeats everything twice.. Also for the other-projects section. hm..

emmahogberg88 (commented):
As I thought I didn't find much to help you with regarding the accessibility. You have used several tools for making the site accessible and that shows off! Really good job Tiina (or Tajna, as the screen reader pronounce it ;) ) ❤️ I did also notice a difference in space between the icons in the more info section that I think is just a "quick fix" 😄

PR #181, opened 2022-03-12
Lovisas portfolio.

LovisaBrorson:
https://lovisas-portfolio.netlify.app/


qwer-ctrl (commented):
Easy and clean code, very good structure. Your css was simple and you have also pay attention for assecibility and I like this yellow color theme. You have achieve all requirements what we have given on figma file.

PR #184, opened 2022-09-16
My portfolio

CeciliaMichele:


styles.css
```diff
@@ -0,0 +1,280 @@
...
+    font-family: 'Montserrat', sans-serif;
+    font-style: normal;
+    font-weight: 400;
+    font-size: 30px;
+    line-height: 50px;
+    margin-left: 5%;
+    margin-right: 5%;
+}
+
+/*Tech-section*/
```
marwebdesign, line 108: These titles makes the css file a lot more readable, good thinking!

index.html
```diff
@@ -0,0 +1,174 @@
...
+<!-- Section with information about me -->
+
+    <section class="about-me" aria-label="Information about me" tabindex="4">
+        <p class="text">
+        <span class="line"></span><span class="dot"></span>
+            Hello! Im an aspiring frontend developer with 8 years of experience within office administration and with a degree in journalism, and a bachelors degree in culture. Autumn of 2021 I decided that I wanted to take my very mixed background into tech and started to learn how to code. My best soft skills are that I'm paying a lot of attention to details and that I'm ambitious. 
+        </p>
+    </section>
+
+    <section class="tech" aria-label="Section about technologies" tabindex="5">
```
marwebdesign, line 49: Regarding tabindex = from what I have read one should not use higher indexes than 0, and instead focusing on making the website readable through logic code. I think your structure is overall good, so I think you don't have to use the tabindex in this way.

index.html
```diff
@@ -0,0 +1,174 @@
...
+        <p class="text">
+            <b>Want to read my mind?</b>
+        </p>
+            <p>Stay tuned for more posts!</p>
+        </div>
+    </section>
+
+<!--My skills-->
+
+    <section class="skills" aria-label="My skills">
```
marwebdesign, line 237: Great job with the aria-labels! Really opened my eyes for what it does for the accessibility when using a screen reader.

index.html
```diff
@@ -0,0 +1,174 @@
...
+        <a href="https://github.com/CeciliaMichele">
+        <i class="fa-brands fa-github-square fa-2xl" aria-label="My Github"></i>
+        </a>
+        <a href="https://www.linkedin.com/in/ceciliaraden/">
+        <i class="fa-brands fa-linkedin fa-2xl" aria-label="My Linkedin"></i>
+        </a>
+
+        <span class="sr-only" tabindex="1">Github</span>
+        <span class="sr-only" tabindex="2">Linkedin</span>
+        </div>
```
marwebdesign, line 30: Both tab-ing and screen reader had some problems with these links, I guess it is because the tabindex is outside of the link class?

marwebdesign (commented):
Great job this week! A lot of things you managed to do. The website looks overall like the design sketch - some work needs to be done with the responsiveness (how to show the content in the different browser-sizes)and the for-more-section. I am really impressed that you get all your links to work (I struggled with that). 

When it comes to accessibility - the screen reader worked really well. You have used the aria-labels in a way that really helps the reader to navigate through the site. It hade some issues with the SoMe-links though. The same when trying to tab through your website (see comment in code for what I think is the issue). 

When running the site through a web accessibility tool, I got som errors and alerts that should be quite easy to fix! Some contrast issues, but also regarding the tab-index (which I also commented on in your code). 

Keep up the good work!

PR #187, opened 2022-09-17
#4 Portfolio- Amanda Elvkull

AmandaElvkull:
https://amandaelvkull-portfolio.netlify.app/


index.html
```diff
@@ -0,0 +1,374 @@
...
+    </div>
+</section>
+  
+<!--Contact-->
+<div class="contact">
+        <h2 class="contact-header">contact</h2>
+        <h2 class="contact-info"> Amanda Elvkull</h2>
+        <h2 class="contact-info">+47 411 13 07</h2>
+        <h2 class="contact-info"> Amanda.elvkull@gmail.com</h2>
+    </div>
```
KajaWilbik, line 370: tips: div in lines 365 and 370 should be in the same line.

style.css
```diff
@@ -0,0 +1,618 @@
+/* Header https://colorhunt.co/palette/820000b9005bff7c7cfee0c0*/
+p {
+    font-family:"Roboto", sans-serif;
+}
+h3 {
+    font-family:'montserrat', sans-serif;
+}
+
+div,
+label {
```
KajaWilbik, line 10: is this styling for both div and label? If yes, maybe it would be better to put them in the same line to make them more readable, like you have in line 83

style.css
```diff
@@ -0,0 +1,618 @@
...
+
+}
+
+.linkedin1 , .github1 , .stackoverflow1 {
+    color:white;
+    
+}
+
+.profil-img img {
+    display: flex;
```
KajaWilbik, line 89: is it display flex necessary here? I tried in the dev tool in chrome to take it out and it looks like it is not changing anything. Have you tried this?

style.css
```diff
@@ -0,0 +1,618 @@
...
+    grid-column: 2 / 4;
+    margin-left:10px;
+    margin-right: 10px;
+}
+
+/*featured projects- grid*/
+.image ,.image-article {
+	display: block; 
+	width: 100%;
+    height: 100%;
```
KajaWilbik, line 156: tip: lines 154 and 155 should be in the same position from the edge as line 156

style.css
```diff
@@ -0,0 +1,618 @@
...
+    display: block;
+    text-align: center;
+    background-color: rgb(185, 0, 91);
+}
+
+.logos-bottom {
+	display: flex;
+	flex-direction: row;
+	justify-content: center;
+    margin-top: 50px;
```
KajaWilbik, line 366: same situation as above, all lines looks better when they are in the same position

KajaWilbik (approved):
Nice and clean code, with small esthetic details (see comments). 
The website looks really nice on the desktop. I love your color design. Using the dev tool I can see that it is a responsive website. Good job. 
Screen reader reads only the title, name, and links. I could not make it to read text. It is how you wanted to work?

AmandaElvkull:
closed

PR #192, opened 2022-09-18
Saralie Bognandi Portfolio

Scarylie:


code/index.html
```diff
@@ -0,0 +1,356 @@
...
+              </h3>
+              <h1 tabindex="0">frontend web developer & design engineer</h1>
+            </div>
+
+            <!-- Profile picture-->
+
+            <img
+              tabindex="0"
+              class="profile-img"
+              src="./pic/sar.bild.jpg"
```
juliaostedt: Recommendation to name your images/icons/assets per the kebab-case style and perhaps slightly more descriptive. Perhaps something like "profile-img-saralie". And for the icons you could for example name them "icon-gh-white" / "icon-li-white" etc. - then they will all be placed next to eachother in the folder and make it easier to view.

code/index.html
```diff
@@ -0,0 +1,356 @@
...
+            </p>
+          </div>
+        </section>
+
+        <!-- PEOJECT SECTION-->
+        <section class="featured-projects block-section">
+          <div class="inner-wrapper">
+            <h2 tabindex="0" class="section-heading">Featured projects</h2>
+            <div class="project-list">
+              <!-- WEATHER APP project-->
```
juliaostedt, line 135: Great use of the comments to structure your code!

code/index.html
```diff
@@ -0,0 +1,356 @@
...
+                  href="https://github.com/Scarylie"
+                  tabindex="0"
+                  role="button"
+                  aria-pressed="false"
+                  aria-label="github"
+                  target="_blank"
+                  rel="noreferrer"
+                >
+                  <img tabindex="0" alt="Github logo" src="./pic/GHwhite.svg" />
+                </a>
```
juliaostedt: With tabindex added to both the a-tag and image it makes the user tab twice to go to next icon. I believe the a href-tag is automatically being focused and a tabindex is therefore not needed. See this article re which HTML-elements that are, by default, focusable: https://bitsofco.de/how-and-when-to-use-the-tabindex-attribute/

code/style.css
```diff
@@ -0,0 +1,376 @@
...
+}
+
+.project-section .content {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  padding: 25px;
+  border-left: 1px solid rgba(0, 0, 0, 0.2);
+  border-right: 1px solid rgba(0, 0, 0, 0.2);
+  border-bottom: 1px solid rgba(0, 0, 0, 0.2);
```
juliaostedt, line 235: I believe you could refactor this by writing:
 ```
 border: 1px solid rgba(0, 0, 0, 0.2);
 border-top: 0;
```
  
  Perhaps worth testing! 😄

File: code/style.css [hunk 5]
```diff
@@ -0,0 +1,376 @@
...
+  h1 {
+    font-weight: 700;
+    font-size: 40px;
+    line-height: 40px;
+  }
+  .profile-img {
+    width: 180px;
+    position: absolute; /* lets us choose position of the img */
+    right: -2vh;
+    top: 22vh;
```
juliaostedt: Your profile img seem to be a bit too far down in chrome, for me it's now showing behind the h4 intro text. Could perhaps try and lower the top value?

juliaostedt (commented):
Good job! Well done with the use of grid/flex and the responsiveness (minor detail with your profile picture but rest looks great.) 👍 😄

PR #193, opened 2022-09-18
Niclas Jensen's Portfolio Site

jnbjensen:


karlfredrikholm:
Looks good overall. Well put together with the choice of colors etc. Not sure how it affects the accessibility, but either way I would've preferred plain text instead of the logos of all the tech stuff. Clean HTML and CSS. Checked it on my mobile too, and it looks great there as well. Good job Nick! 👍

PR #194, opened 2022-09-18
project portfolio

hannanorgard:


AgnesAhlman (commented):
If I understood correctly from your read me file you are aware of the issue with the mobile-version, but besides from that it looks great!

PR #205, opened 2022-09-18
Tara project-portfolio

Tara78:


Earnieball:
It looks and functions great, the only thing I had to comment on was the color-scheme. Maybe use some more subtle colors? But I guess that's an subjective viewpoint=) Anyways, really good job!

PR #206, opened 2022-09-18
JessMo-Project portfolio 

jessnmo:


style.css
```diff
@@ -0,0 +1,390 @@
...
+        padding: 10px;
+    }
+
+    .thought-section{
+        display: flex;
+        flex-direction: column;
+        height: fit-content;
+        overflow: hidden;
+    }
+}
```
Bibiruu, line 390: I like your color palette ideas. 

Also once you portfolio is done , if you would  like to have a smooth scroll effect for it to "slide" when checking out your page, you can use the 
html {
    scroll-behavior: smooth;
  }

for this. Good job of how you built this in a week even if you are busy with your work =)

style.css
```diff
@@ -0,0 +1,390 @@
...
+    background-color:rgb(157,163,157);
+    border-radius: 5px;
+}
+
+.project-section{
+    display: grid;
+    width: 100vw;
+    height: 150vh;
+    grid-template-areas: 
+    "title title title title"
```
Bibiruu, line 130: I like how the layout turns out with appointing the areas. Gives a quick picture of the styling. =)

index.html
```diff
@@ -0,0 +1,224 @@
+<!DOCTYPE html>
```
Bibiruu, line 1: Super creative the icon on the upper tab beside your name!

PR #208, opened 2022-09-18
My portfolio - Jessica Hansson 

jesshansson:
https://jessica-hansson.netlify.app/


index.html
```diff
@@ -0,0 +1,252 @@
...
+            </a>
+        </div>
+        <img class="profile-pic" src="Pictures/jhansson.png" alt="Profile picture of Jessica Hansson">
+    </header>
+
+    <!--Intro section--> 
+        <section class="intro-section">
+            <p>
+            <span class="line"></span>
+            <span class="dot"></span>
```
jessikalind: I found that if you move the two `<span>`s to one line, the space between the line and the dot disappears 😃

index.html
```diff
@@ -0,0 +1,252 @@
...
+
+    <!--Intro section--> 
+        <section class="intro-section">
+            <p>
+            <span class="line"></span>
+            <span class="dot"></span> 
+            Coming from a background within biology and environmental science, I have now taken on the challange to become a frontend developer. Combining my experience from project and business development, I hope to grow a career within programming and contribute to the progress of the tech world. </p>
+        </section>
+
+    <!--Tech section-->
```
jessikalind, line 51: Nice use of comments throughout the HTML file. It makes it nice and easy to get a quick overview :)

style.css
```diff
@@ -0,0 +1,523 @@
...
+    color: black;
+    cursor: pointer;
+    text-align: center;
+}
+
+.intro-section, .tech-section, .featured-projects, .other-projects, .skills-section, .for-more, footer {
+    display: flex;
+    flex-direction: column;
+    align-items: center; 
+}
```
jessikalind, line 18: Very good that you combined the code for the different sections. Spares a lot of lines!

style.css
```diff
@@ -0,0 +1,523 @@
...
+header {
+    display: flex;
+    flex-direction: row;
+    color: white;
+    height: 520px;
+    background-color: #483643; 
+    background-image: url("Pictures/background1.jpg");
+    background-repeat: no-repeat;
+    background-size: cover; 
+}
```
jessikalind, line 87: An idea is to use an inner wrapper for the heading as well, so that the width could be set to 80% here as well (since the content in the header was aligned with the content in the body in the design that we were given)

style.css
```diff
@@ -0,0 +1,523 @@
...
+    font-style: normal;
+    font-weight: 400;
+    font-size: 17px;
+    line-height: 29px; 
+    padding: 65px 24px 35px 24px;
+    background-color: #F4F4F4;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+}
```
jessikalind, line 162: Also here, an option could have been to use an outer- and inner wrapper. The same goes for the Tech section. Perhaps that would have been a way to make the content more aligned throughout the page ☀️

style.css
```diff
@@ -0,0 +1,523 @@
...
+    border-radius: 5px;
+}
+
+.featured-projects {
+    font-family: 'Montserrat';
+    font-style: normal;
+    font-weight: 400;
+    font-size: 17px;
+    line-height: 25px;
+}
```
jessikalind, line 225: Perhaps by making .featured project a flex (or grid) parent, the wrap-function could have worked (ie displaying 2x2 projects in desktop view). I used CSS grid, and the code (snippet) for mobile first looked like this: 
```
.featured-projects {
    display: grid;
    grid-template-columns: 1fr;
```
For the larger screens it however looked like this:
```
.featured-projects {
    grid-template-columns: 1fr 1fr;
    column-gap: 3vw;
}

    .one {
        grid-row: 2 / 3; 
    }

    .two {
        grid-row: 2 / 3;
    }

    .three {
        grid-row: 3 /4;
    }

    .four {
        grid-row: 3 / 4;
    }
}
```

index.html
```diff
@@ -0,0 +1,252 @@
...
+        </section>
+
+
+    <!-- Footer -->
+    <footer>
+        <div class="contact-container">
+            <h2 class="title-contact">CONTACT</h2>
+            <p class="contact-info">Jessica Hansson</p>
+            <p>+46 73 511 77 16</p>
+            <a href="mailto:jessica.hansson@hotmail.se" target="_blank">jessica.hansson@hotmail.se</a>
```
jessikalind, line 251: Very nice that you made the email adress into a link - I forgot about that!

index.html
```diff
@@ -0,0 +1,252 @@
...
+                            <h3 class="date-header">SEP 2022</h3>
+                            <h4 class="blog-headline">The start of a journey!</h4>
+                            <p class="blog-post">Lorem ipsum dolor sit amet consectetur adipisicing elit. Quasi, eaque?</p>
+                        </div>
+                    </div>
+
+            <!--More thoughts section--> 
+                <div class="more-thoughts">
+                    <h2 class="small-h2">MORE THOUGHTS</h2>
+                        <h4 class="blog-headline">Stay tuned!</h4>
```
jessikalind, line 183: Semantically it might be better to jump from an h2 to an h3 (instead of h4).

index.html
```diff
@@ -0,0 +1,252 @@
...
+        @import url('https://fonts.googleapis.com/css2?family=Montserrat&display=swap');
+    </style>
+    <link rel="icon" type="image/png" sizes="32x32" href="Pictures/favicon-32x32.png">
+</head>
+<body>
+    <header class="header">
+
+        <div class="header-name-title">
+            <h1 tabindex="0">PORTFOLIO:<span class="jessica-name-header">JESSICA HANSSON</span></h1>
+            <h2>frontend <br> developer</h2>
```
jessikalind: To make it a bit smoother for the screen reader, I have learnt it is good to avoid using <br>. I solved it by limiting the width of the "headline"-component.

jessikalind (commented):
You've done a great job on your portfolio, Jessica! (Even though I know you have told me that you are not fully finished yet). It looks close to the design we are given, and it works fine with screen reader tools ☀️

Some general thoughts are that you could have expanded your use of outer-and inner div:s, to make the whole page align better (see my more detailed comments below). I have also made a suggestion on how you could make the featured projects align in two rows when on desktop-view. 

When looking at your site through the Wave-tool, it looks pretty good! One of the things it showed is however that there are some contrast issues on the site (I have them too), but that is easily fixed by changing the color of some of the headings. 

Thank you for giving me an insight in your educational and professional background - it all sounds so interesting 🐯😍

index.html
```diff
@@ -0,0 +1,252 @@
...
+                    </a> 
+                </div>
+
+                <div class="project two">
+                    <a href="https://jessica-linda-weather-site.netlify.app/" target="_blank">
+                        <div class = "project-img-container">
+                            <div class="project-img-overlay">
+                                <p aria-hidden="true">WEATHER APP</p>
+                            </div>
+                            <img src="Pictures/weatherapps.JPG" alt="Weather app">
```
jessikalind: Maybe this alt text is not needed (since the img is mostly decorational, and the name of the site is read by the screen reader in the h3 below. This goes for all the featured projects 😇

PR #212, opened 2022-09-18
David Ballester - Project Portfolio

Fonnt:
https://davidballester.dev/


code/index.html
```diff
@@ -0,0 +1,361 @@
...
+        <p>PORTFOLIO: <span class="bold">DAVID BALLESTER</span></p>
+        <h1>frontend developer</h1>
+        <h2>+ leading</h2>
+      </div>  
+    </div>
+  </header>
+<!-- INTRO SECTION -->
+  <section class="outer-wrapper-grey">
+    <div class="inner-wrapper-grey">
+      <p class="paragraph-style" tabindex="0">
```
ElinSegelov: I assume that you intentionally wrote this part in Swedish. Regardless of what language you use I think maybe you should be consistent and use the same language all over the site.

code/index.html
```diff
@@ -0,0 +1,361 @@
...
+  <section class="outer-wrapper-grey">
+    <div class="inner-wrapper-grey">    
+      <h3 class="section-titles-blue" tabindex="0">FEATURED PROJECTS</h3>
+
+      <div class="featured-projects-wrapper">        
+<!-- CARD -->
+        <div class="featured-projects-card">
+          <a href="https://idyllic-queijadas-6e6760.netlify.app/" target="_blank">
+            <div class="featured-projects-image">
+              <img class="card-image" src="./assets/weather-app.jpg" alt="">
```
ElinSegelov: make sure to and an alt-ext for all images

code/index.html
```diff
@@ -0,0 +1,361 @@
...
+    <div class="inner-wrapper-grey">    
+      <h3 class="section-titles-blue" tabindex="0">FEATURED PROJECTS</h3>
+
+      <div class="featured-projects-wrapper">        
+<!-- CARD -->
+        <div class="featured-projects-card">
+          <a href="https://idyllic-queijadas-6e6760.netlify.app/" target="_blank">
+            <div class="featured-projects-image">
+              <img class="card-image" src="./assets/weather-app.jpg" alt="">          
+              <h4 class="centered" aria-hidden="true">WEATHER APP</h4>
```
ElinSegelov: It's good that you hide this. Gets really repetitive with the screen reader otherwise.

code/index.html
```diff
@@ -0,0 +1,361 @@
...
+                <h5>JavaScript ES6</h5>
+                <h5>APIs</h5>
+              </div>
+            </div>
+          </a>    
+          <a href="https://github.com/Fonnt/project-weather-app" target="_blank">
+            <div class="tag-wrapper-deployment">
+              <h5>GitHub Repo</h5>
+            </div>
+          </a>
```
ElinSegelov: Great idea adding this here!

code/index.html
```diff
@@ -0,0 +1,361 @@
...
+      <div class="footer-wrapper">
+        <div class="footer-card">
+          <div class="footer-info">            
+            <div class="footer-paragraph-style">
+              <p tabindex="0"><span class="bold">CONTACT</span></p>
+              <p tabindex="0">David Ballester</p>
+              <p><a class="address-style" href="tel:+46708363717">+46 7 08 36 37 17</a></p>
+              <address>
+                <a class="address-style" href="mailto:dballesterfont@gmail.com">
+                  e-mail
```
ElinSegelov: For consistency maybe type out your email-address here too. And a suggestion is to add some text for the screen reader clarifying that this is your phone number and e-mailadress. I did it using a screen reader only class, hiding the content from view. If you google "screen reader only class" you will get some tips on that css properties you might need.

code/style.css
```diff
@@ -0,0 +1,961 @@
...
+  top: 75%;
+
+  width: 150px;
+  height: 150px;
+  
+  border-radius: 50%;
+  
+  border: 2px solid whitesmoke;
+  
+  overflow: hidden; /* Makes sure nothing sticks out of the circle */
```
ElinSegelov, line 135: maybe remove the empty lines

code/style.css
```diff
@@ -0,0 +1,961 @@
...
+
+  color: #000000;
+}
+
+.section-titles-blue {
+  margin: 5vh auto;
+  padding: 0.5vh 0.5vh;
+ 
+  margin-bottom: 0;
+  font-family: 'Montserrat';
```
ElinSegelov, line 202: Instead of specifying font-family for every section, maybe set a family for the entire document and only change where needed.

code/style.css
```diff
@@ -0,0 +1,961 @@
...
+  width: 100%;
+  height: 200px;
+  filter: brightness(75%);
+  object-fit: cover;
+}
+
+/* --------------MY THOUGHTS-------------- */
+/* --------------INFO PART-------------- */
+
+.my-thougts-info{
```
ElinSegelov: This has alot of the same styling as other-projects-info. Could they be using the sam class styling and then just change the specific properties where they need to be different? In this case, bottom margin.

code/style.css
```diff
@@ -0,0 +1,961 @@
...
+  width: 10px;
+  background-color: #406882;
+  /* background-color: #FA382F; */
+  border-radius: 50%;
+  display: inline-block;
+}
+
+/* --------------TAGS-------------- */
+
+.tag-wrapper {
```
ElinSegelov, line 563: If you add `flex-wrap: wrap;` I think it will take care of the fact that JavaScript ES6 is in two lines under one of your projects.

ElinSegelov (commented):
First of all, I think the site looks great! 
I really like the color-gradiant of the header background as well as you choice of color palette.
You have consistent indentationing and have structured your code well. Your css file is very large though and there are alot of styling that you are repeating so I think it would be great if you looked over your code and see how you could make styled classes that work for several sections. This might bread up the structure you have now, sorting the styling under different sections and so on. You also have alot of empty lines within the styling for various elements. I don't know if that is intentional? For me however, I think it's easier to get a good overview if all the styling for a specific element is just stacked up on each other with no lines separating it. I also this that is considered best practice but can remember where I've read it.

Great job David!

PR #213, opened 2022-09-18
Project-portfolio

YoTesfam:
https://younas-tesfamariam-portfolio.netlify.app/


Code/index.html
```diff
@@ -0,0 +1,234 @@
...
+	<!--Header and Top Class-->
+	<header class="header-section">
+		<div class="header-container">
+			<div class="social-media-icons-header">
+				<a href="https://www.aftonbladet.se/">
+					<img class="social-icon" src="./assets/linkedin.svg" alt="LinkedIn logo"></a>
+				<a href="https://github.com/YoTesfam">
+					<img class="social-icon" src="./assets/github.svg" alt="Github logo">
+				</a>
+
```
Fonnt, line 23: It would be easier to follow your code if you change the indentation in your vscode to 2 spaces instead of 4 spaces. Doing so, you could get consistency and the code wouldn't go so deep into the center of the page. :) You can also change and save those settings so every time you open vscode you get the same indentation. Easy fix. I can show you if you don't find it

Code/index.html
```diff
@@ -0,0 +1,234 @@
...
+			<div class="header-text-group">
+				<p class="header-portfolio"> PORTFOLIO:
+					<span class="header-name">YOUNAS TESFAMARIAM</span>
+				</p>
+				<h1>Frontend developer</h1>
+
+			</div>
+		</div>
+	</header>
+	<main>
```
Fonnt, line 35: it's fine to have the main tag but I think remember that is and old practice. I think it can be omitted and use `<section>` tag instead

Code/index.html
```diff
@@ -0,0 +1,234 @@
...
+		<section class="section">
+			<div class="container">
+				<h2>Tech</h2>
+				<p class="tech-text">
+					HTML, CSS, Git, JavaScript
+				</p>
+			</div>
+		</section>
+
+		<div class="projects">
```
Fonnt, line 57: this `<div>` might be turned into a `<section tag>`

Code/styles.css
```diff
@@ -0,0 +1,469 @@
+body {
+  margin: 0;
+  box-sizing: border-box;
+  font-family: 'Montserrat', sans-serif;
+}
+
+.header-section {
```
Fonnt, line 7: Your text info is flying around when you try the responsivness. You could add a position relative to the section and then position absolut to the diferents elements. Then you could use % or vw or vh for example in order to position those elements as you wished

Code/styles.css
```diff
@@ -0,0 +1,469 @@
...
+}
+
+.intro-text {
+  width: 80%;
+  max-width: 700px;
+  position: relative;
+  height: 300px;
+  margin: 40px auto;
+  line-height: 1.5;
+}
```
Fonnt, line 87: You can also achieve that positioning, creating a conatiner with display flex, wrap row or column and telling the child to act the way you like

Code/styles.css
```diff
@@ -0,0 +1,469 @@
...
+
+.section:nth-child(2n) {
+  background-color: #f4f4f4;
+}
+
+
+
+
```
Fonnt, line 118: empty lines ;)

Code/index.html
```diff
@@ -0,0 +1,234 @@
...
+			<div class="projects-section">
+				<div class="grid-item">
+					<div class="image">
+						<img class="project-image" src="./assets/guess.who2.png" alt="Guess who app screenshot"
+							width="360" height="250">
+					</div>
+					<div class="project-text">
+						<h3>Guess who</h3>
+						<p>Guess the right character by asking questions.
+							<a href="https://sprightly-nougat-44de5e.netlify.app/"><span class="arrow">>></span></a>
```
Fonnt, line 70: The link is only active if you hover this : >>

You can try to wrap the hole div with the `<a>` so you can click over the hole project. More user friendly

Code/styles.css
```diff
@@ -0,0 +1,469 @@
+body {
+  margin: 0;
+  box-sizing: border-box;
+  font-family: 'Montserrat', sans-serif;
+}
+
```
Fonnt, line 6: I think you could create an outerwrapper and an innerwrapper, give them the values that you want and then repeat them into every section. Doing that you could be able to fix the margins that apeear when trying different responsivness and you could also control easier the positioning of your elements.

Code/styles.css
```diff
@@ -0,0 +1,469 @@
...
+.grid-item {
+
+  border: solid 3px rgba(197, 230, 197, 0.527);
+  font-size: 19px;
+  text-align: center;
+  line-height: 1.5;
+  margin: auto;
+}
+
+@media (min-width: 381px) and (max-width: 767px) {
```
Fonnt, line 254: I think you could change the media queries here for tablets into `@media (min-width: 668px) and (max-width: 1023px) `
and another for desktop like this `@media (min-width: 1024px) {`

Fonnt (commented):
Hi Younas! Great work with the portfolio! It wasn't easy to manage to achieve all the goals! I think you did a really good job! You have a little responsivness issue that can be fixed with outer and inner wrappers and  then using flexbox. You could also try to fix the media queries into the correct pixel size. Maybe it would fix it. 

Try to work a little bit more with proper indentation and removing empty lines. It can easly be done changing the value in the vscode from 4 to 2 spaces. Doing that it would be easier to follow yourt code.

I like the colour palette you used. Looks nice! 
I would also try to change the links and make the hole div clickable, not only the >> signs.

Remember that you can use tabindex  in the html in order to improve keyboard navigation. It helps you also to make a more user friendly webpage.

Well done! Keep doing and reach me out if you want to discuss something!

/David Ballester

PR #214, opened 2022-09-18
Project Portfolio

Kristin-Larsson:


Kristin-Larsson:
https://jovial-concha-89db34.netlify.app/

index.html
```diff
@@ -0,0 +1,205 @@
...
+        <icon class="icons">
+            <a class="linkedin-link" href="https://www.linkedin.com/in/kristin-larsson-83a559113/" target="_blank">
+            <img class= "linkedin" src="./icons/linkedin.svg" alt="Link to Linkedin"></a>
+            <a class="github-link" href="https://github.com/Kristin-Larsson" target="_blank">
+            <img class= "github" src="./icons/github.svg" alt="Link to Github"></a>
+            <a class="stackoverflow-link" href="https://stackoverflow.com/c/technigo/users/395" target="_blank">
+            <img class= "stackoverflow" src="./icons/stackoverflow.svg" alt="Link to Stackoverflow"></a>
+        </icon>
+<!-- Hero-text -->
+        <span class="hero-text">
```
petrasson, line 28: I think span is an inline element and can't be in a block element (h1 e.g) So maybe use divs instead of span in some cases?

index.html
```diff
@@ -0,0 +1,205 @@
...
+<section class="intro">
+        <p><span class="line"></span><span class="dot"></span>
+            Lorem ipsum dolor sit, amet consectetur adipisicing elit. Corporis id, adipisci, dolore et pariatur velit minima nisi autem, cupiditate ex magnam magni accusantium ut quisquam ratione commodi modi mollitia dicta.
+        </p>
+</section>
+
+<!--------------- Tech --------------------------->
+<section class="tech">
+    <span class="title">TECH</span>
+    <section class="tech-container">
```
petrasson, line 46: Perhaps use div instead of section here so you have only one section tag?

index.html
```diff
@@ -0,0 +1,205 @@
...
+            <div class="program-container">
+                <p class="programs">HTML</p> <p class="programs">CSS3</p> <p class="programs">JavaScript ES6</p>
+            </div>
+        </div>
+    </section>
+</section>
+    
+<!---------------- Other projects------------------->
+<section class="other-projects">
+    <span class="otherprojects-title">OTHER PROJECTS</span>
```
petrasson, line 83: perhaps you can optimize your code by using the same class for titles, text or boxes that are the same? for example can you use the same class here as the title "more thoughts"?

petrasson:
Nice one Kristin!
I really like the colors and design. The page works fine, is responsive, and looks good in different browsers.
Perhaps just go through the margins as well as the errors in validator.w3.org.
You have used a lot of semantic elements in your code but I saw that some of the errors are related to that, so solving them will make it even better from an accessibility perspective! 
Well done!

PR #217, opened 2022-09-18
Kim Svensson - Portfolio

Earnieball:


jnbjensen (commented):
Veery late code review, and fairly redundant now that we're redoing our portfolio sites with React anyway, but: Great work as always! Only thing I felt was 'missing' was some kind of hover effect to the social media buttons. : )

PR #219, opened 2022-09-18
Sofia Andersson - Project Portfolio

Sofia-Andersson:
Week 4


index.html
```diff
@@ -0,0 +1,294 @@
...
+                <img src="./IMAGES/stackoverflow-icon.png" alt="stack-overflow-icon">
+            </a>
+        </section>
+
+        <section class="headings">
+            <h3 class="heading1">PORTFOLIO: <span class="name">SOFIA ANDERSSON</span></h3>
+            <h1 class="heading2">frontend <br> developer</h1><br>
+            <h2 class="heading3">+ digital designer</h2></span>
+        </section>
+            <img class="header-image" src="./IMAGES/BildSofia.jpg" width="180" height="180" alt="Accessible image Sofia">
```
katirum, line 34: I like how you specify the name for this picture so it's accesible. However I wonder why you modify the height and width in html instead of css?

index.html
```diff
@@ -0,0 +1,294 @@
...
+
+    <section class="featured-projects-container">
+        <h4>FEATURED PROJECTS</h4>
+
+        <div class="project-container">
+            <a class="project-card" href="https://github.com/Sofia-Andersson/project-guess-who">
+              <div class="project-1-image">
+                <p class="project-title">GUESS WHO GAME</p>
+              </div>
+              <section class="project-description" aria-label="Project description">
```
katirum, line 63: Nice use of aria-label. I have to start using that as well!

index.html
```diff
@@ -0,0 +1,294 @@
...
+            Description project 4.</span>
+          <span class="link-arrows" aria-hidden="true">>></span>
+        </section>
+        <section class="code-tag-container" aria-label="Code tags">
+          <div class="code-tag">JavaScript ES6</div>
+          <div class="code-tag">React.js</div>
+          <div class="code-tag">Styled Components</div>
+        </section>
+      </a>
+
```
katirum, line 155: a neat trick to put this whole thing in section

index.html
```diff
@@ -0,0 +1,294 @@
...
+            <h6 class="list-title">CODE</h6>
+            <ul title="CODE">
+              <li>HTML5</li>
+              <li>CSS3</li>
+              <li>JavaScript ES6</li>
+              <li>React.js</li>
+              <li>Git</li>
+            </ul>
+          </div>
+
```
katirum, line 226: I never thought about using list for this, but it's cool that it's also working this way!

style.css
```diff
@@ -0,0 +1,663 @@
...
+.list-title {
+  font-size: 17px;
+}
+
+/*HEADER-CONTAINER*/
+
+.social-media-links{
+  position: absolute;
+  display: flex;
+  align-items: flex-start;
```
katirum, line 97: I wonder why you use display flex when you already use display block with position absolute?

katirum (commented):
Overall, I think it's a neat project. Good work, Sofia! :)

index.html
Sofia-Andersson: I tried specifying both in css & html and for some reason I couldn't make it work in css. But I agree it should be possible. :)

style.css
Sofia-Andersson: From what I understand display (flex) and position are not doing really related to each other and both can be used on the same item. I would explain it as position is connected to how items react when changing circumstances: size of screen etc. while flex is more about how items are placed in relation to its flex-friends. :) (google probably can give you better explanations.)

Anyway if I remove display-flex in this case, the icons will be placed in a vertical row instead of horizontal.

PR #220, opened 2022-09-18
Portfolio by Linnéa Ajger

LinneaAjger:


reset.css
```diff
@@ -0,0 +1,43 @@
...
+figure, figcaption, footer, header, hgroup, 
+menu, nav, output, ruby, section, summary,
+time, mark, audio, video {
+	margin: 0;
+	padding: 0;
+	border: 0;
+	font-size: 100%;
+	font: inherit;
+	vertical-align: baseline;
+}
```
CeciliaMichele, line 20: Really smart way to gather everything like this!

index.html
```diff
@@ -0,0 +1,126 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="reset.css">
+    <link rel="stylesheet" href="style.css">
+    <title>Linnéa's portfolio</title>
+</head>
+<body>
+    <header></header>
```
CeciliaMichele, line 12: The end tag is right after the opening tag for the header

style.css
```diff
@@ -0,0 +1,407 @@
...
+    text-align: center;
+}
+.skills p {
+    font-family: 'Roboto', sans-serif;
+}
+
+.for-more {
+    display: flex;
+    flex-direction: column;
+    align-items:center;
```
CeciliaMichele, line 221: You have a few of those align-items where there's no space in between the : and the attribute

CeciliaMichele (commented):
A really nice looking portfolio with really good contrasts for accessability, which also really make it look really nice with all the black/white/purple. It sure looks like the original but with a personal touch to it. The links works perfectly fine and it's all very large and easy to read and and it looks very professional. The background picture in the header is not following when you go from mobile to bigger screens, I think that's because you have the opening and closing tag at the same place. It was also fun to see that you used the input wrapper! I think the code is very consistent, with the exception that there's a few lines where it should be a blank space between the css class and the attribute, but otherwise it looks really nice. Really good job this week, I think you have a really good ground to continue to build on!

PR #223, opened 2022-09-18
Neah Rockstroh - portfolio

Archkrull:


LisaVasic:
The code is very clean and easy to understand. As I can see there are no duplicates. Good use of comments!

The site works in Chrome, Firefox as well and Edge.
It also works with screen reader and keyboard navigation. 

A suggestion to make it more accessible would be a bit better contrast in the header-text.

I like the use of inner and outer wrapper and wish I would have done the same. 

Really nice job!

style.css
```diff
@@ -0,0 +1,499 @@
+@import url('https://fonts.googleapis.com/css2?family=Montserrat:ital,wght@0,400;0,700;1,400&family=Roboto:ital@0;1&display=swap');
+
+
+:root {
+    --primary: #f8f8f8;
+    /* background colour 1 */
+    --secondary: white;
```
LisaVasic, line 7: This is really cool and something I will use in future projects.

PR #226, opened 2022-09-18
Portfolio Arnau Monforte Coronas

Iracuzzi:
https://arnaumonforte.netlify.app/


code/style.css
```diff
@@ -0,0 +1,278 @@
...
+    flex-wrap: wrap;
+    padding: 5px;
+    margin-bottom: 20px;
+}
+
+
+.project {
+    border: solid 2px black;
+    padding: 10px;
+    background-color: rgb(249, 236, 236);
```
jessnmo, line 134: Clean code very easy to navigate. But for this section maybe give each project div a height & width so the block size will stay the same

code/style.css
```diff
@@ -0,0 +1,278 @@
...
+    padding: 40px;
+    align-content: stretch;
+    justify-content: space-evenly;
+    flex-wrap: nowrap;
+    align-items: center;
+}
+
+.container h2 {
+    display: flex;
+    flex-direction: row;
```
jessnmo, line 203: the default flex direction is row so unless you want the direction as column you don't have to specify

PR #228, opened 2022-09-18
Linus Erikssons portfolio

erilinus:


project 1.html
```diff
@@ -0,0 +1,49 @@
...
+                        <li><a href="#">About</a></li>
+                        <li><a href="#">Contact</a></li>
+                    </ul>
+                </nav>
+            </div> -->
+    <nav class="nav-parent">
+        <a class="nav-child home"href="index.html">Linus</a>
+        <ul class="nav-child">
+            <li><a href="#">Project 1</a></li>
+            <li><a href="#">Project 2</a></li>
```
SoderstromC: How come I can't find these two Projects on the site? Can only see "About" and "Email me".

SoderstromC (commented):
I like the overall styling and use of a navbar. For me it feels more like a CV than a portfolio and I think that could have been addresses by following the portfolio guidelines below:
The portfolio should contain:
* 		A picture of you.
* 		A presentation of the tech skills that you have.
* 		A presentation of some thoughts that you have around code.
* 		A presentation of your projects linking to your Github.

I works well in several browsers and is responsive. Accessibility is ok but was a bit hard to assess thoroughly due to the page not containing a lot of content.

My recommendation would be to weave your experience into the "About me" paragraph and replace the current Experience section with a Project overview section instead. 
I personally don't agree with the use of a profile picture since it caters to bias in recruitment, so from my point of view you can keep as is and leave it out.

PR #229, opened 2022-09-18
Carina Moreno - Portfolio

carimore:


index.htm
```diff
@@ -0,0 +1,227 @@
...
+        <div class="otherproject-container">
+        <!-- Future Project A -->
+        <article class="futureProject">
+            <a href="https://tourmaline-sable-678f53.netlify.app/"
+                target="blank"
+                class="project-link"
+                areal-label="link to visit my collaborative Weather App Project">
+            <div class="project-name">
+            </div>
+            </a>
```
uohman: The anchor tag is now empty. The closing tag should maybe be moved to row 157 instead? :)

index.htm
```diff
@@ -0,0 +1,227 @@
...
+            </article>
+        <!-- Future Project B -->
+        <article class="futureProjectB">
+            <a href="https://soft-tiramisu-65c499.netlify.app/"
+                target="blank"
+                class="project-link"
+                areal-label="link to visit my Future Project">
+            <div class="project-name">
+            </div>
+            </a>
```
uohman, line 170: Same as comment above. Maybe move the closing anchor tag to row 173?

index.htm
```diff
@@ -0,0 +1,227 @@
+<!DOCTYPE html>
+<html lang="eng">
```
uohman: When going through the code in a validation tool it says: "The language subtag eng is not a valid ISO language part of a language tag.". Maybe try with "en" instead of "eng"?

index.htm
```diff
@@ -0,0 +1,227 @@
...
+                target="blank"
+                class="project-link"
+                areal-label="link to visit my collaborative Weather App Project">
+            <div class="project-name">
+                <div class="overlay"> WEATHER APP </div>
+                <img class="images" src="Images/weatherAppProject.png" alt="Weather App Project"/>
+            </div>
+            </a>
+            <h3>WEATHER APP</h3>
+            <p class="descriptions">This was a collaborative project to make a Weather App using API</p2>
```
uohman: Why p2 in the end tag?

index.htm
```diff
@@ -0,0 +1,227 @@
...
+        <!-- Future Project A -->
+        <article class="futureProject">
+            <a href="https://tourmaline-sable-678f53.netlify.app/"
+                target="blank"
+                class="project-link"
+                areal-label="link to visit my collaborative Weather App Project">
+            <div class="project-name">
+            </div>
+            </a>
+            <h3>WEATHER APP</h3><p class="descriptions">This was a collaborative project to make a Weather App using API</p2>
```
uohman, line 155: Why p2 in the end tag?

index.htm
```diff
@@ -0,0 +1,227 @@
...
+            <div class="tag-wrapperother">
+                <p>HTML5</p>
+                <p>CSS3</p>
+                <p>JavaScript</p>
+                <p>REACT</p>
+            </div>
+            </article>
+        </div> <!-- div for project-container-->
+
+        </div> <!-- div for otherproject-container-->
```
uohman, line 181: To where does this end tag div belong?

uohman (commented):
Overall a nice start with the project. :) Thought I won't comment on the CSS since you're still working on that. I used [https://wave.webaim.org/](url) and [https://validator.w3.org/](url) to go through your html and they gave a little bit different alerts. I think it would be good to look through the page especially with [https://validator.w3.org/](url) to get rid of some errors already now though you aren't finished.

But a good start and especially nice work with the header! It looks very nice! :)

PR #233, opened 2022-09-18
Caroline Soderstrom Portfolio

SoderstromC:
https://cosmic-treacle-0e00c7.netlify.app/


style.css
```diff
@@ -0,0 +1,457 @@
+:root {
+    /* ---primary:  #F2A2E4; another option for primary */
+    ---primary:  #eb72d5d1;
+    ---secondary: #d4c4af;
+    ---main: #dfd3c38a;
+    ---highlight: #F090D9;
+    ---accent: #C400FF;
+}
```
a-well, line 8: Nice work on the colours! The splashes of pink and the colourful pins in the header look great together with the other more neutral colours

style.css
```diff
@@ -0,0 +1,457 @@
...
+.footer-container {
+    padding: 0px;
+    }
+
+.footer-heading {
+    font-size: 20px;
+}
+
+.footer-box {
+    width: fit-content;
```
a-well, line 438: I noticed that the text in the footer wasn't being centered on mobile, and removing this line seemed to do the trick 🙂

a-well (commented):
Great job! The project meets all the requirements and that while looking fantastic ✨

General feedback:
* consider adding a max width to the `.container` class to not make it stretch on really wide screens

PR #235, opened 2022-09-18
Portfolio, Nina Berggren

NinaBerggren:
Link to netlify:https://portfolio-nina-berggren.netlify.app/


code/index.html
```diff
@@ -0,0 +1,326 @@
...
+            <a class="icon-circle" title="link to Nina's Github page" href="https://github.com/NinaBerggren" aria-label="Go to Github page">
+            <i aria-hidden="true" class="fab fa-github"></i>
+            </a>
+        </div>
+        <div class="heading-container" tabindex="0">
+            <h1 class="heading-1 uppercase">portfolio: <span class="bold">NINA BERGGREN</span></h1>
+            <h1 class="heading-2">frontend</h1>
+            <h1 class="heading-2">developer</h1>
+            <h1 class="heading-3">+ problem solver &amp;
+            <h1 class="heading-3 indent"> structured "doer" </h1>
```
Archkrull: In every section, to properly follow the syntax there's only supposed to be one h1, one h1, on h3 and so forth. So for accessibilitys sake it would be better to have everything in the same h1 or make the other ones into <p>

code/style.css
```diff
@@ -0,0 +1,603 @@
+@import url('https://fonts.googleapis.com/css2?family=Montserrat:wght@400;700&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Roboto:wght@400;700&display=swap');
```
Archkrull, line 2: Just a tip, is is possible to import several fonts in the same import URL!

code/style.css
```diff
@@ -0,0 +1,603 @@
...
+
+/* ----------COLOR SCHEME---------- */
+:root {
```
Archkrull, line 5: Great with the root! Makes it a lot easier to follow. You can also add font-size in root

Archkrull (commented):
Amazing that you managed this over a weekend! Most things I'll comment on you'd definetely have done if you had more time I'd imagine.
First of all, great that you use root, it makes it mush simpler to change the design. For future reference I'd style the font sizes with root as well and then em to make it more responsive and easier to update.
Generally I'd also rather style margins, padding and similar with %, vh, vw etc rather than pixels because there is such a huge variety of sizes, so % makes it easier to design responsive sites.

You got the structure in, and also started with accessibilty so I think you should be proud with what you managed in just a weekend! When you have time make it work for screen readers as well. Good job!

PR #237, opened 2022-10-23
Project Portfolio- Fanny Cardell

fannycardell:


index.html
```diff
@@ -0,0 +1,325 @@
...
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+<link href="https://fonts.googleapis.com/css2?family=Montserrat:wght@400;700&family=Roboto:wght@400;700&display=swap" rel="stylesheet">
+</head>
+<body>
+    <header class="heading-container">
+        <div class="inner-wrapper">
+            <div class="heading-icons">
+                <a href= "https://www.linkedin.com/in/fanny-cardell-354a4863/">
+                <img src="./images/linkedin-icon.png" alt="Linkedin icon">
```
LinneaAjger: Looks like the icons don't exist in the images folder 🤷🏻‍♀️

index.html
```diff
@@ -0,0 +1,325 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./reset.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+<link href="https://fonts.googleapis.com/css2?family=Montserrat:wght@400;700&family=Roboto:wght@400;700&display=swap" rel="stylesheet">
+</head>
+<body>
+    <header class="heading-container">
+        <div class="inner-wrapper">
```
LinneaAjger, line 15: Looking at this element in the Inspector, it's just laying on the top. I think you can remove it, or if you want the header content to have the same margins you can just add: margin: 0 auto; width= 80% (or how much you want on the sides). And you should add position: relative, so that all your children-elements with position:absolute is referring to the inner-wrapper :)

style.css
```diff
@@ -0,0 +1,510 @@
...
+    display: inline-block;
+    padding: 20px;
+}
+
+.featured-project-flex-container {
+    position: relative;
+    display: flex;
+    flex-wrap: wrap;
+    width: 100%;
+    justify-content: center;
```
LinneaAjger: Maybe `space-evenly` can give the projects some more space, so they don't scale over each other when hovering. :)

style.css
```diff
@@ -0,0 +1,510 @@
...
+    margin: 10px;
+    /* border: purple 2px solid;  */
+}
+
+ .featured-project-wrapper {
+    position: relative;
+    flex-direction: row;   
+    width: 90%;
+    height: 450px;
```
LinneaAjger: When i scale down the screen your project images gets more thin. Maybe this is like you want it, but I think you can fix so that the proportions are the same even when scaling down =)

style.css
```diff
@@ -0,0 +1,510 @@
...
+    }
+    .featured-project-wrapper {
+    width: 500px;
+    height: 450px;
+    }    
+
+    .skills-grid-container{
+    grid-template-columns: repeat(3, 1fr);
+    } 
+}
```
LinneaAjger: Are you thinking of changing the font sizes when changing from mobile to desktop for example?

index.html
```diff
@@ -0,0 +1,325 @@
...
+            <div class="skills-grid-container">
+                <div class="skills-wrap-order">
+                <h6>Code</h6>
+                    <ul class="skill-grid-code">
+                        <li class="skill-list">HTML5</li>
+                        <li class="skill-list">CSS3</li>
+                        <li class="skill-list">JavaScript ES6</li>
+                        <li class="skill-list">GitHub</li>
+                        <li class="skill-list">React</li>
+                        <li class="skill-list">Styled Components</li>
```
LinneaAjger, line 369: Nice work with the grid element! 🌟

index.html
```diff
@@ -0,0 +1,325 @@
...
+         <!-- Footer-->
+         
+    <footer>
+        <div class="footer-container">
+            <div class="inner-wrapper"> 
+                <h2 class="footer-heading">CONTACT</h2>
+                <div class="footer-info">
+                    <p>Fanny Cardell</p>
+                    <p>+46 733 46 22 34</p>
+                    <a href="mailto:fannycardell@gmail.com">fannycardell@gmail.com </a>
```
LinneaAjger, line 423: Nice detail with the mail-link! You can add the similar one to your phone number as well

index.html
```diff
@@ -0,0 +1,325 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./reset.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+<link href="https://fonts.googleapis.com/css2?family=Montserrat:wght@400;700&family=Roboto:wght@400;700&display=swap" rel="stylesheet">
```
LinneaAjger, line 11: I think you're missing the tag <title> here!

LinneaAjger (commented):
Well done Fanny!! Your portfolio looks nice and all the major functions seem to work! I like your header and that you've used a lot of color, makes it stand out! I can see that you've work both with flexbox and with CSS grid and that's cool that you show off both skills. Your page also works great to tab through and with the screen reader. But when I tested your site in the validation tools, there was some problems. Have a look at:
https://validator.w3.org/
https://www.accessibilitychecker.org/
I am also missing the styling for desktop screen.

Fix those things all your project will meet the requirements! 🌟

PR #246, opened 2023-03-04
Leo Thunell portfolio

khleoth:


noraterbocs (commented):
You have done a really good job. I can see that your project looks quite similar to the provided design and your html looks organised. 
Regarding the improvements, I have noticed that the margins in the tablet view are not wide enough and the featured project section pushes the page to the side. 
All in all, I think you are learning quite fast and you are on a good way to become a great frontend developer. :)

PR #253, opened 2023-03-05
Project-Portfolio/EvaSjöholm

EvaSjoholm:
A portfolio-project with HTML and CSS that should also be added Accessibility to it. There is a lot left to do and I will finish the project when learning more about correct code for links as well as CSS design. 

Responsive page to all devices, this will need to be checked and corrected because lack of access of differents devices. My version is a small desktop so I worked my best with that. 

Live link:
https://zippy-dieffenbachia-054803.netlify.app/


code/index.html
```diff
@@ -0,0 +1,341 @@
...
+            </svg>
+          Live demo   
+         </a>
+    
+          <a href="https://github.com/EvaSjoholm/project-guess-who" class="git-code-button" target="_blank">
```
jenster5, line 116: This svg tag is really bulky and takes up a ton of space in your code you can get rid of it by --- first  adding the fontawesome stylesheet to your HTML by either downloading the stylesheet or use a CDN link. Next have it referenced and styled in the css and then  apply the CSS class to the element you want to include the icon.  there is an example here-----

```
<a href="https://github.com/EvaSjoholm/project-guess-who" class="git-code-button" target="_blank">
  <i class="icon fa fa-github"></i> Github
</a>
```

code/index.html
```diff
@@ -0,0 +1,341 @@
...
+        </svg>
+      </a>
+       </div>
+</section>
+  <script src="script.js"></script>
```
jenster5, line 339: if you dont use js can you remove this? it gives an error in the console with the js doc not found

jenster5 (commented):
Hey I think your page is lovely and you did a fantastic job! I just had a few comments on things you can fix if you want but honestly I think its nice as it is also. First was only if you wanted the info for my profile pic when I wanted to make it round mine also turned into an oval shape instead of circle and daniel said it was because my starting image was not square. So I made my starting image square and got the circle in the design instead of the oval. But that is very design related maybe you liked the oval better just an fyi. The only other comments are the functionality in different sized in the mobile version for the featured projects one button is yellow but the others are not. The tablet view you need more space between the columns in Skills, the my words buttons one is yellow and the hover effects dont work for me in tablet but they do in desktop. 

Hope these comments help a bit and overall really great job on a really tough project!

PR #256, opened 2023-03-05
Project portfolio

EmiliaGA:


index.html
```diff
@@ -0,0 +1,238 @@
...
+                <a href="https://www.linkedin.com/in/emiliaga/">
+                    <img src="linkedin.svg" alt="linkedin link">
+                </a>
+                <a href="https://github.com/EmiliaGA">
+                    <img src="github.dark.svg" alt="github link">
+                </a>
+            </div>
+        </div>
+        <div class="scroller">
+            <img src="scroller-salmon.png" alt="down arrow" class="scroller">
```
SandraMadeleine: Nicely done with the arrow and text!

index.html
```diff
@@ -0,0 +1,238 @@
...
+            <h1>Tech</h1>
+        </div>
+        <div class="h-skills">
+            <p>HTML, CSS, Flexbox, <span class="js-mark">JavaScript,</span> Web Accessibly, API:s, 
+                <span class="mp-mark">mob-programming,</span> pair-programming, Github. </p>
+        </div>
+    </section>
+
+    <section class="projects">
+        <h1>Featured projects</h1>
```
SandraMadeleine: This entire section looks really good! It's responsive and has the buttons with hover, positions changing etc. Good job!

index.html
```diff
@@ -0,0 +1,238 @@
...
+    <title>Emilias Portfolio🌈</title>
+</head>
+
+
+<body>
+
+<header>
+    
+    <section class="top-container">
+        <div class="picture-big">
```
SandraMadeleine: It's a really nice and welcoming image, and the colors in the entire profile are well matched with this initial picture of the sky. :-)

SandraMadeleine (commented):
It seems that the css-file is too big to open here and when I open it in a separate tab, I can not comment. Not sure how to fix go around that unfortunately. But I looked at your css and you have done a massive job, really good work getting the grid to work so nicely everywhere, and the buttons too! Thumbs up! From what I can tell things work as they should. The skills list is a bit wonky in some sizes, but I couldn't see anything in the code that I could easily point to as the culprit.

PR #258, opened 2023-03-05
week 4 - portfolio - Edvard Ekström

edv-rd:


EmiliaGA (changes requested):
Neat and tidy code. Simple to follow and the comments for each section makes it easier to follow. I'm missing a link to the deployed site in the read-me, perhaps something you've forgotten? :) 
Otherwise well done, as always!

PR #262, opened 2023-03-05
Project portfolio

AnnikaSonnek:


malvinamaria (commented):
Overall structure: The HTML code is structured well and follows the standard conventions. The use of semantic tags such as header, section and div is good for readability and maintainability

index.html
```diff
@@ -0,0 +1,412 @@
...
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta property ="og:title" content="Annika's Portfolio"/>
+    <meta property="og-url" content="https://annikasonnek.netlify.app/"/>
+    <meta property="og:type" content="website"/>
+    <meta property ="og:description" content="The portfolio page of Annika Sonnek frontend developer student at Technigo."/>
+    <meta property ="og:image" content="https://https://annikasonnek.netlify.app/images/pictureofme2.jpg"/>
```
malvinamaria: In the `meta` tag with `property`="og:image", the `content` attribute has a duplicated `https:// `in the URL:

style.css
```diff
@@ -0,0 +1,732 @@
...
+  body {
+    overflow-x: hidden;
+  }
+  
+  body {
+    position: relative;
+  }
+  
+  /*FONT STYLING*/
+  .paragraph {
```
malvinamaria, line 17: The` font-family` property for the `.paragraph `selector is missing a fallback font. It's best practice to provide multiple font options, in case the first choice is not available on the user's computer.

style.css
```diff
@@ -0,0 +1,732 @@
...
+    color: white;
+    padding: 0px 2px 0px 2px
+  }
+
+  .dark-blue {
+    background-color: #113f8f;
+    color: white;
+    padding: 0px 2px 0px 2px
+  }
+
```
malvinamaria, line 115: The `.green`, `.dark-green`,` .light-blue`, and .`dark-blue `classes have the same padding values. It might be a good idea to extract this to a separate class, and apply it to all four selectors, to reduce repetition.

style.css
```diff
@@ -0,0 +1,732 @@
...
+    font-style: normal;
+    font-weight: 700;
+    font-size: 80px;
+    line-height: 80px;
+    color: #40513B;
+    display: flex;
+    justify-content: center;
+  }
+
+  .decription-header{
```
malvinamaria, line 73: There's a typo in the `.decription-header `selector - it should be .de**s**cription-header.

malvinamaria (commented):
It's good to see using `alt` attribute for images and using semantic tags for better screen reader support. You could add more descriptive text to the images and providing a skip link for keyboard-only users.

PR #263, opened 2023-03-05
Project Portfolio

camcron:


index.html
```diff
@@ -0,0 +1,382 @@
...
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+        <link rel="stylesheet" href="./css/reset.css">
+        <link rel="stylesheet" href="./css/style.css">
+        <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=EB+Garamond">
+
+        <meta property="og:title" content="Camilla's portfolio">
+        <meta property="og:description" content="The portfolio page of Camilla Cronqvist, frontend developer">
+        <meta property="og:image" content="./assets/pexels-negative-space-92904.jpg">
```
mvfrid, line 14: The og tag image is not showing, since it's not an absolute URL (we point to a picture in our local folder). A tip is to check what it will look like here: https://www.opengraph.xyz/url/https%3A%2F%2Fportfolio-camilla-cronqvist.netlify.app%2F

index.html
```diff
@@ -0,0 +1,382 @@
...
+        <meta property="og:image:alt" content="Picture of computer with code">
```
mvfrid, line 15: Good to include the alt tag also in the og section!

index.html
```diff
@@ -0,0 +1,382 @@
...
+
+        <link rel="icon" type="image/png" href="./assets/icons8-portfolio-50.png">
```
mvfrid, line 17: Ooh this looks very nice! Great job!

index.html
```diff
@@ -0,0 +1,382 @@
...
+                            xmlns="http://www.w3.org/2000/svg"
+                        >
+                            <path
+                                d="M10.1875 25V10.9844H5.82812V25H10.1875ZM7.98438 9.10938C9.39062 9.10938 10.5156 7.9375 10.5156 6.53125C10.5156 5.17188 9.39062 4.04688 7.98438 4.04688C6.625 4.04688 5.5 5.17188 5.5 6.53125C5.5 7.9375 6.625 9.10938 7.98438 9.10938ZM26.4531 25H26.5V17.3125C26.5 13.5625 25.6562 10.6562 21.25 10.6562C19.1406 10.6562 17.7344 11.8281 17.125 12.9062H17.0781V10.9844H12.9062V25H17.2656V18.0625C17.2656 16.2344 17.5938 14.5 19.8438 14.5C22.0938 14.5 22.1406 16.5625 22.1406 18.2031V25H26.4531Z"
+                                fill="currentColor"
+                            />
+                        </svg>
+                    </a>
+                    
+                    <a href="https://github.com/camcron" role="button" title="Link to GitHub" class="icons" target="_blank" aria-label="Link to Camilla's GitHub profile">
```
mvfrid, line 63: Really good job making the icons work for screen readers! Very clear description of them as well.

index.html
```diff
@@ -0,0 +1,382 @@
...
+                            fill="currentColor"
+                            />
+                        </svg>
+                    </a>
+                </div>
+
+                <div class="scroll" aria-label="Scroll down to see more content">
```
mvfrid, line 79: Same here with the screen reader accessibiility. Very nice and clear description.

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+    src: url('../fonts/fonts/Satoshi-Variable.woff2') format('woff2'), url('../fonts/fonts/Satoshi-Variable.woff') format('woff'), url('../fonts/fonts/Satoshi-Variable.ttf') format('truetype');
+    font-weight: 300 900;
+    font-display: swap;
+    font-style: normal;
+}
+
+* {
+    box-sizing: border-box;
+}
+
```
mvfrid, line 15: I would write a little comment that shows that the css below is more "general" and applies to elements all over the portfolio. So it's clear for the reader where to look for formatting that might affect several areas.

E.g.:
/*------  GENERAL FORMATTING -----*/

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+.main-info-footer h4 {
+    font-style: normal;
+    font-weight: 900;
+    font-size: 32px;
+    line-height: 43px;
+    margin: 0;
+}
+
+.personal-photo {
+    margin-bottom: 35px;
```
mvfrid, line 448: 1) This applies to the image in both the footer and the header, if you want to target only the footer, then write `.footer-box .personal-photo`.

2) I would recommend checking out the flexbox "gap" property. In many cases it's better, instead of using margin to create a space between flex-box elements. If you use gap, then you don't need to move the margin- (i.e. from margin-bottom to margin-left) when you switch between column and row in different queries.  Gap works for both! It really helped me simplify my own portfolio code :)

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+}
+
+.icons-box-footer {
+    grid-column: 2 / span 2;
+    display: flex;
+    justify-content: space-around;
+    align-self: flex-start;
+}
+/*FOOTER MOBILE END*/
+/*END MOBILE*/
```
mvfrid, line 469: Great use of comments to mark where the section also ends!

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+    color: #FFFFFF;
+}
+
+/*HEADER MOBILE START*/
+header {
+    display: grid;
+    grid-template-columns: repeat(4, 1fr);
+    grid-template-rows: 1fr 1fr 30px 225px;
+    padding: 55px 24px;
+    height: 920px;
```
mvfrid: Try setting it to a min-height instead and allow it to grow if the screen is small, if you want to have the scroller on the white part of the section. Otherwise it's ok as it is!

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+    align-items: flex-start; 
+}
+
+footer h2 {
+    grid-column: span 4;
+    display: flex;
+     font-style: normal;
+     font-weight: 900;
+     font-size: 54px;
+     line-height: 162px;
```
mvfrid, line 412: Is this from Figma? It looks a bit strange with the large line-height and the margin-top for footer-box. Could we decrease the line-height and add more margin instead? Would make it easier to follow in chrome inspector.

index.html
```diff
@@ -0,0 +1,382 @@
...
+                        </p>
+
+                        <div class="code-tags">
+                            <p>HTML5</p>
+                            <p>CSS3</p>
+                            <p>JavasScript</p>
+                        </div>
+
+                        <div class="project-buttons">
+                            <a href="https://github.com/camcron/project-chatbot" role="button" class="code-button" title="Link to GitHub" target="_blank">
```
mvfrid, line 678: Screen reader does not read any of these buttons (git & live) on my phone

index.html
```diff
@@ -0,0 +1,382 @@
...
+                    , React Hooks, Redux, Node.js, Mongo DB, Web Accessibly, API:s,
+                    <span class="mob-techstack">mob-programming</span>
+                    , pair-programming, Github.
+                </p>
+            </section>
+
+            <section class="projects">
+                <h2>Featured Projects</h2>
+
+                <article class="project-chat">
```
mvfrid, line 660: To make the code more recyclable and scalable (when adding more cool stuff to the portfolio), I would recommend using generic classes for all articles, i.e. "project-article" or just "article" instead of "project-chat". The styling will be the exact same even though the text content (set in HTML) is different. If we need to style images differently, add a specific class to that perticular image.

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+    font-family: 'Satoshi-Variable';
+    font-style: normal;
+    font-weight: 500;
+    font-size: 16px;
+    line-height: 22px;
+}
+
+.code-tags {
+    display: flex;
+}
```
mvfrid, line 327: Duplicate to row 308-310?

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+
+.code-tags p {
+    color: #ffffff;
+    background-color: #000000;
+    margin: 4px;
+    padding: 2px 6px;
+    font-size: 16px;
+}
+/*PROJECTS SECTIONS MOBILE END*/
```
mvfrid: Duplicates to row 312-322?

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+.live-button img {
+    width: 48px;
+    height: 48px;
+}
+
+.live-button {
+    width: 164px;
+    height: 48px;
+    margin-bottom: 80px;
+}
```
mvfrid, line 119: Are we using this to write over the properties set in row 88-106 or is it an accidental double?

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+
+
+/*PROJECTS SECTION MOBILE START*/
+.projects {
+    display: grid;
+    grid-template-columns: repeat(4, minmax(0, 1fr));
+    grid-template-rows: auto;
+    justify-content: center;
+    justify-items: center;
+    align-items: flex-start;
```
mvfrid, line 268: We don't seem to use display: flex here, so these would not do anything

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+article {
+    display: flex;
+    flex-direction: column;
+    grid-column: 2 / span 2;
+    justify-content: center;
+}
+
+article:last-child {
+    margin-bottom: 200px;
+}
```
mvfrid, line 289: Nice solution!

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+
+.project-info h3 {
+    font-family: 'Satoshi-Variable';
+    font-style: normal;
+    font-weight: 900;
+    font-size: 32px;
+    line-height: 43px;
+    margin-bottom: 12px;
```
mvfrid, line 297: Just a suggestion, it would look nice to have some space towards the project image. I would maybe add a margin-top of 10px or similar. Or add a flexbox gap on the "project-chat" class.

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+    display: flex;
+    justify-content: space-around;
+    align-self: flex-start;
+}
+/*FOOTER MOBILE END*/
+/*END MOBILE*/
+
+
+/*MEDIA QUERY TABLET*/
+ @media (min-width: 667px) {
```
mvfrid, line 473: I always write the tablet query like this: `@media (min-width: 667px) and (max-width: 1024px) {}`

Good to know this also works :)

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+    /*HEADER TABLET START*/
+    header {
+        display: grid;
+        grid-template-columns: minmax(102px, 1fr) repeat(6, 77px) minmax(102px, 1fr);
+        grid-template-rows: 412px 244px 180px 32px 1fr;
+        height: 920px;
+        align-items: center;
+        padding: 0;
+    }
```
mvfrid, line 482: We have a lot of repetitions used here that is not really needed, since they are already defined in the mobile version. For instance: "display: grid" and "height: 920px". If you have time later maybe it could be good to go through the code and clean out these duplicates, to make it smaller and more "clean".

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+    .header-box {
+        grid-area: 2 / 2 / 3 / 8;
+        display: flex;
+        flex-direction: row;
+        margin: 90px 0 20px 0;
+        width: auto;
+    }
+
+    .photo-box {
+        margin-right: 40px;
```
mvfrid, line 502: Here I again suggest flexbox gap :)

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+    align-items: center;
+    justify-content: center;
+    color: #232c39;
+    background-color:  #F5F5F5;
+    border-radius: 40px;
+    border: none;
+    margin-top: 24px;
+    margin-right: 24px;
+    font-size: 18px;
+    padding: 0;
```
mvfrid, line 101: The button content is not fully centered within the button. It's not a big deal and still looks really good, but if you want to fix that then I would add padding-left and padding-right.

css/style.css
```diff
@@ -0,0 +1,919 @@
...
+    }
+
+    .footer-box {
+        grid-area: 2 / 2 / 3 / 8;
+        display: flex;
+        margin-top: 74px;
+    }
+
+    .main-info-footer {
+        text-align: left;
```
mvfrid, line 907: I'm pretty sure this is default? It did not change anything when I removed it at least.

index.html
```diff
@@ -0,0 +1,382 @@
...
+        <meta property="og:description" content="The portfolio page of Camilla Cronqvist, frontend developer">
+        <meta property="og:image" content="./assets/pexels-negative-space-92904.jpg">
+        <meta property="og:image:alt" content="Picture of computer with code">
+
+        <link rel="icon" type="image/png" href="./assets/icons8-portfolio-50.png">
+        <title>Camilla's Portfolio</title>
+    </head>
+
+    <main>
+        <body>
```
mvfrid: I get a few errors in the validator when I run the code. I think it has to do with the <body> being inside <main>, which should be the other way around.

https://validator.w3.org/nu/?doc=https%3A%2F%2Fportfolio-camilla-cronqvist.netlify.app%2F

index.html
```diff
@@ -0,0 +1,382 @@
...
+            <header>
+                <div class="side-box-image" aria-hidden="true">
+                    <p>Only for tablet and desktop purpose</p>
+                </div>
+
+                <div class="header-box">
+                    <div class="photo-box">
```
mvfrid, line 29: Is the div photo-box really neccessary here? What does it do? Can we not add a margin directly to the img "personal-photo"?

mvfrid (approved):
**Overall feedback:**

- The design is super nice and clean, looks professional. Good job selecting colors, images and sizes.

- We commented on this already on the demo, but nice formatting of the code. You are very good with adding proper indentation, making it easier to look at and follow.

- Look at the fonts to make sure they are behaving as they should, when I unclick stuff like "font-weight" and "type" for some elements, nothing happens. Maybe it's just my browser? Not sure..

- Responsiveness is also really good, everything looks good in all the sizes I could test. In tablet, we have some fixed sizes that could be changed to have it behave similarly as mobile/desktop. But it's not really a problem, it still looks good even though it's a bit static.

- Great work on the validation, solving all the accessibility issues with nearly no warnings/errors.

- I'm not sure about the approach with the many columns and rows. It would be great if you could ask that on SO and get the code coaches view on if that's how we should do it. Would be interesting to know! It feels very robust :) 

General advice to go over the code and structure it up a bit more when you have time. There are a lot of properties that I'm not sure you really need. It makes it easier to understand the differences between the media queries and some elements used in several places, and easier for you to add to the portfolio later. Use variables if you have some properties that you are re-using in more than a few places.


**Summary:**
Really good job! It is functional and looks very good in all forms, you should give yourself a pat on the back for this weeks work!

PR #265, opened 2023-03-05
Project Portfolio

sammyolsson:


IrupePozoGraviz (approved):
Yey! Good job it looks super good and its easy to use works with screenreader, a few problems that I saw were the colors on the Tech part I know they were directly from figma but could be good to change them to make it more readable

When I look at the page in size 1028 -1267 the first picture seems utdragen, and I think the recommendation is to already design for start desktop 1024 or 1025 (thats what Tecnigo said in the precourse I think). And last the scroll thing on tablet has a tendency to lay upon the words some time could be good to lok at that.
I think its a awesome job and what you have been able to do in this short time!

PR #266, opened 2023-03-05
Portfolio-project still in progress

cecavi:


code/index.html
```diff
@@ -0,0 +1,309 @@
...
+
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link href="https://fonts.googleapis.com/css2?family=EB+Garamond&display=swap" rel="stylesheet">
+</head>
+
+<body>
+
+    <div class="introduction-side-image">
+        <span class="sr-only">Background image with colored pencils</span>
```
EmmaHoltegaard, line 27: Great that you use an "sr-only".

code/index.html
```diff
@@ -0,0 +1,309 @@
...
+             </a>
+
+            </div>
+        </div>
+    </section>
+
+
+
+    <section class="tech">
+        <h1>Tech</h1>
```
EmmaHoltegaard, line 85: I was told at some point, that ` <h1>`  should only be used once, and always for the first thing on the page (in this case, your name). I think it's a screen reader-thing, and not everyone seems to agree.

code/index.html
```diff
@@ -0,0 +1,309 @@
...
+            </div>
+          </div>
+
+      </section>
+
+      <section class="skills">
+        <h1>Skills</h1>
+
+        <div class="skill-lists-container">
+            <dl class="skill-lists code">
```
EmmaHoltegaard, line 211: I've never seen a dl before, only ul or ol. Cool :)

code/css/style.css
```diff
@@ -0,0 +1,659 @@
...
+    src: url('./Satoshi_Complete/Fonts/WEB/fonts/Satoshi-Variable.ttf') format('truetype');
+  }
+
+  @font-face {
+  font-family: SatoshiBold;
+  font-weight: bold;
+  src: url("./Satoshi_Complete/Fonts/WEB/fonts/Satoshi-Bold.ttf");
+}
+
+  :root {
```
EmmaHoltegaard, line 21: I've seen a few people do colours like this, and It seems like a really good idea. So much easier to change the colors later on!

EmmaHoltegaard (approved):
Really good work. The site works without any problems, as far as I can tell, and looks very close to design brief (especially desktop). In terms of the design, only a few things jumped out. On tablet, the skills-section should be in columns of 2, and in the featured-projects section (tablet), the widths on images and text a little tweaking. Otherwise, it looks really good.

PR #267, opened 2023-03-05
Daniel Brobäcks Portfolio

dannebrob:
https://technigo-portfolio.netlify.app/


cpz-crafts (commented):
Hi Daniel!! first of all Amazing job! 🥳  the responsiveness of the website is really good and overall you followed very well the design given which is impressive. I really love the scrolling effect , the logo of the front page!. something to add is that the in the skills  part the tittles  have the default font so maybe is worth it to add the satoshi font there and maybe give it a little bit more of margin top and bottom  in the desktop version . but overall wow! really nice portfolio 🥇

PR #270, opened 2023-03-05
project-portfolio Fanny

fannystenberg:


code/style.css
```diff
@@ -303,6 +302,20 @@ footer h2 {
     margin-left: 24px;
 }
 
+/* RESPONSIVE DESIGN FOR SMALL MOBILES */
```
fikamate, line 351: Nice!!

code/index.html
```diff
@@ -52,18 +52,18 @@ <h2 class="title">Frontend developer</h2>
...
+            <!-- Here goes "Lets scroll" -->
+    <div class="scroll" aria-hidden="true">
+        <p class="arrow"><i class="fa-solid fa-arrow-down"></i></p>
+        <p class="scroll-text">Lets scroll</p>
+    </div>
 
     <main>
          <!-- Here goes info about tech-stack -->
-        <section class="tech">
+        <section class="tech" title="Information about tech-stack" aria-label="Information about tech-stack">
```
fikamate: Im not 100% sure but it might be "kaka på kaka" using both title and aria-label attribute, since both labels fill the same function for screen readers. Double check this, i might be wrong. But if its the case i think you should remove the aria-label since that is considered being the last resort if theres no other way to provide accessible labels for an element.

fikamate (commented):
REALLY good work Fanny!! Im super impressed with your work with this, the code is very clear and structured and easy to follow. Great that you tested your site for screen readers and  included styling for small devices also. All i could see as improvement points from my (very limited hehe) perspective was some minor things you might wanna check up about the use of aria labels. At some places they might be unnecessary to have since the html markup is semantically correct and the tags and written content gives a clear information about the content. But thats just to stretch your goals :) Great job!!

PR #273, opened 2023-03-05
Project Portfolio

smirrebinx:
View it Live
https://smirrebinx-project-portfolio.netlify.app/


majazimnoch:
Hi! It's hard for me to comment on every new commit branch as there is so much code this time!
I'm going through your code in chrome dev tools and it's easier to catch some things. Så nu kör vi:
1. I see the icons in header-section are wrapping inside the div when changing the web size. As a k fix, I can suggest adding some top padding to the class icon (at least 130px) so it looks ok on all media. 
2. The picture for the tablet section - I think you need to experiment with the object-fit property for the image in the project section to look as on the design.
3. I like that the buttons have shadows! 
4. I think all sections should have the same top and bottom padding.
5. The photo of you - I think the circle is not really round - I have the same problem and I don't know yet how to fix it.
6. I like the light, mint colors. They work calmly for my eyes! 
7. Well done! Almost finished and I see that your web has no errors in the Markup Validation Service so good job! The version for the mobile looks almost finished (maybe apart from the last section) so I can imagine you were working first on mobile, then tablets, and then desktop. Maybe it would be worth continuing somehow on three of them at the same time next time? ;D

PR #275, opened 2023-03-05
Nina Waldréus Portfolio

NinaWald:
https://nina.waldreus.se/


css/satoshi.css
```diff
@@ -0,0 +1,178 @@
...
+ * Satoshi Italic
+ * Satoshi Medium
+ * Satoshi Medium Italic
+ * Satoshi Bold
+ * Satoshi Bold Italic
+ * Satoshi Black
+ * Satoshi Black Italic
+ *
+*/
+
```
EvaSjoholm, line 24: If this aren't codes to used, perhaps avoiding to keep theme in the css file to create more space and "cleaning up" as they say =)

index.html
```diff
@@ -0,0 +1,312 @@
...
+    
+    
+    <title>Portfolio</title>
+</head>
+<body>
+    <!--This is the main header with image and name info -->
+    
+   
+    <header>
+
```
EvaSjoholm, line 26: Maybe avoiding these spaces between to get more clarity which code is together or a part of it's sections =)

index.html
```diff
@@ -0,0 +1,312 @@
...
+                        Previous to her pharmacist career she studied art in highschool.
+                        She creates user-friendly applications and has a creative and colourful soul. 
+                        Her immence background and drive for excellence makes her a valuable asset to any project 
+                        and a standout in the technology field.
+                        </article>
+                    
+
+                            <div class="icon-container">
+                                <ul>
+                                    <li><a target="_blank" href="https://www.linkedin.com/in/nina-waldr%C3%A9us-b46abb104/"
```
EvaSjoholm, line 53: Perhaps having each code that is together, on one line rather than breaking it into two?

EvaSjoholm (commented):
Very beautiful page, love the color choices and how personal it became. 
Regarding the code - a little bit of cleaning up in the html file, keeping codes that are within it's closing-tag on one line, try avoiding more space than necessary in the file so it's faster to read and follow. 

Was not sure about the  font css./satoshi.css file? Seems like a lot of code written for something that maybe can be solved an easier way perhaps? Just minor thoughts, nothing big. Overall great job. 

I saw that the social-media icons worked so well done! And the hover-effects to. 
Good job Nina!

PR #276, opened 2023-03-05
Tess Aquilon - Project Portfolio

codeandjazz:
https://venerable-starburst-9472e5.netlify.app/


code/reset.css
```diff
@@ -0,0 +1,48 @@
+/* http://meyerweb.com/eric/tools/css/reset/
```
dannebrob, line 1: Nice with a separated reset css file

code/index.html
```diff
@@ -0,0 +1,405 @@
...
+                      </svg>                 
+                      View the code
+                    </div>
+                  </button>
+                </a>
+                <a href="https://fantastic-meringue-7f5119.netlify.app" target="_blank">
+                  <button type="button" class="btn2">
```
dannebrob: Naming classes can be challenging; I found it to be good practice to try not to call things like "btn1", or "btn2". I think that naming more with what they are supposed to be displaying/doing is a more clear way to do it, i.e. "demo-btn" or "yellow-button"

code/style.css
```diff
@@ -0,0 +1,729 @@
...
+    animation: bounce 1s infinite alternate;
+    margin-bottom: 4px;
+}
+
+@keyframes bounce {
+    0%   {transform: translateY(0);}
+    100% {transform: translateY(5px);}
+  }
+
+/* Bounce animation Link: https://codepen.io/technigo/pen/OJoWwbE */
```
dannebrob, line 301: If not intended to use in production, you might want to remove the comment

code/style.css
```diff
@@ -0,0 +1,729 @@
...
+    padding: 20px 20px;
+}
+
+.logo-container {
+    display: none;
+    background: #FCEDEA;
+}
+
+/* Media queries */
+@media screen and (min-width: 350px) and (max-width: 667px) {
```
dannebrob: Since you have the media queries in this excellent structure, you don't need to set the min-width and max-width on every query. You could inherit the styles from the last one. 
Example: 
@media screen and (min-width: 350px) {
.....
}
@media screen and (min-width: 668px) {
.....
}
@media screen and (min-width: 1024px) {
.....
}

dannebrob (commented):
A well made portfolio page. You can be proud! 
The code is well structured and the it shows that you have a good understanding of HTML and CSS. I just added a few tips in my comments.
Great work!

PR #278, opened 2023-03-05
Unfinished

SandraMadeleine:


ninastroyer (commented):
God job Sandra! I know you're not completely finished but you'll get there😁 
The code seems to be well-organized and easy to read, with clear class names and comments. Overall the html looks really good.  Like you say in your Readme the flex is not working, the site isn't responsive and you have to make it accessible but other then that great job👏💫

PR #288, opened 2023-03-06
Jennifer Feenstra

jenster5:
This is only the mobile version 

https://helpful-froyo-056ec5.netlify.app/


code/reset.css
```diff
@@ -0,0 +1,43 @@
...
+del, dfn, em, img, ins, kbd, q, s, samp,
+small, strike, strong, sub, sup, tt, var,
+b, u, i, center,
+dl, dt, dd, ol, ul, li,
+fieldset, form, label, legend,
+table, caption, tbody, tfoot, thead, tr, th, td,
+article, aside, canvas, details, embed, 
+figure, figcaption, footer, header, hgroup, 
+menu, nav, output, ruby, section, summary,
+time, mark, audio, video {
```
khleoth, line 13: Thats alot of things having this applied to them. Would another way be applying things to body at top of CSS?

fonts/css/satoshi.css
```diff
@@ -0,0 +1,178 @@
...
+       url('../fonts/Satoshi-LightItalic.woff') format('woff'),
+       url('../fonts/Satoshi-LightItalic.ttf') format('truetype');
+       font-weight: 300;
+       font-display: swap;
+       font-style: italic;
+}
+
+
+@font-face {
+  font-family: 'Satoshi-Regular';
```
khleoth, line 93: Might want to have a look at what we actually use in the figma? I could count only 3 different satoshi, hence you could reduce the amount of code, unless you want to use all these weight-variants

khleoth (commented):
I guess I mostly have a pointer since you have not had the time that you wished was available past week.

Try going for either grid och flexbox, check the layers in Figma and use that, it helps alot. Apply those divs and get to work. If you structure it with flexbox(or grid) it makes it alot easier modifying with inspect as you go to apply the media queries. Good job scientist, Stay on it!

PR #290, opened 2023-03-07
New pull request

SofiaGerdmar:
Accidentally (!) deleted the repo and had to fork again. Thus a new pull request.


jonsjak (commented):
The overall design looks good. Nice picture and personal styling (scroll icon, text-tags, buttons etc.)

At the moment the page is not responsive which makes it a bit hard to read in tablet and mobile view - I'm guessing due to lack of time.

I noticed that the social media-icons are black as default and grey when hovered over, it's meant to be the other way around. That's an easy fix: just set the color to grey HEX-color from Figma and set the color to black on hover.

The images sits well in the sections, so does the text content.

The links are working fine. If you want them to open in a new tab you can add target="_blank" in the opening <a>-tag. Also if you want, try setting the phone and email in an address-section (here's a link: https://www.w3schools.com/tags/tag_address.asp) and change the email to <a> so it's send an email.

Good job! You have a good base for your future portfolio!

PR #291, opened 2023-03-09
Added text and styling

RegSod:


index.html
```diff
@@ -0,0 +1,252 @@
...
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="style.css">
+    <!-- https://fonts.google.com/specimen/EB+Garamond -->
+    <title>Reginas Portfolio</title>
+</head>
+<body>
+    <!-- Parent -->
+<container class="main-container">
```
AjmalZ, line 13: `<container>` is not a valid HTML tag, I would use `<div>` or `<section>` instead.  

Ex. `<div class="main-container">`

index.html
```diff
@@ -0,0 +1,252 @@
...
+
+
+
+
+ <container class="box1">
+             <div class="box1-1">
+
+             <div class="img-of-me-box"><img class="img-of-me" src="./img/img_reg.jpg" alt=""></div>
```
AjmalZ, line 21: The images in this project is missing in the project folder.

AjmalZ (commented):
There are some tags that needs to be changed and some images that needs to be imported to the project folder. Otherwise it is a good start with the structure. Good job!

PR #308, opened 2023-10-14
Wk8 - Beckie Morton - Project Portfolio

BeckieMorton:
I still have styling to fix, but assuming this will be an ongoing project, I will continue to work on perfecting it.


my-react-portfolio/src/Components/FeaturedProjects/FeaturedProjects.css
```diff
@@ -0,0 +1,76 @@
...
+/*media query for mobile*/
+@media screen and (max-width: 667px) {
+  .featured-projects-container h1 {
+    font-size: 32px;
+  }
+
+  .projects-child {
+    grid-column: span 2;
+  }
+}
```
olgalepisto: Oh I see you have converted your one-file-CSS into multiple files! You seem to have quite a good amount of them, not too many.

my-react-portfolio/src/Components/FeaturedProjects/FeaturedProjects.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+  const formatGitName = (name) => {
+    console.log(`this is in the name:${name}`);
+    let newName = name.split("-").join(" ");
+    console.log(newName);
+
+    const capitals = newName.split(" ");
+    for (var i = 0; i < capitals.length; i++) {
+      capitals[i] = capitals[i].charAt(0).toUpperCase() + capitals[i].slice(1);
+    }
```
olgalepisto: Really nice work with the whole API call and name-formatting thing! (and thanks again, your ideas are present in my code as well....:D)

my-react-portfolio/src/Components/FeaturedProjects/FeaturedProjects.jsx
```diff
@@ -0,0 +1,111 @@
...
+                src="/Assets/featured-projects-temp.jpg"
+              />
+            </div>
+            <div className="projects-child">
+              <h2>{formatGitName(myGit.name)}</h2>
+              <p>
+                {myGit.description}
+                <p>
+                  <span style={{ backgroundColor: "black", color: "white" }}>
+                    &nbsp;HTML5&nbsp;
```
olgalepisto: I'm wondering if it would be better practice to make space around the language name with padding in CSS instead of using &nbsp;? If I remember correctly, using these kinds of HTML elements isn't recommended, but I could be wrong!

my-react-portfolio/src/Components/FeaturedProjects/FeaturedProjects.jsx
```diff
@@ -0,0 +1,111 @@
...
+                  </span>
+                  &nbsp;
+                  <span style={{ backgroundColor: "black", color: "white" }}>
```
olgalepisto: Also here I'm wondering if the styling could exist inside a CSS document instead of using the style tag multiple times with the same content?

my-react-portfolio/src/Components/Footer/Footer.css
```diff
@@ -0,0 +1,37 @@
...
+  white-space: nowrap;
+}
+
+@keyframes ticker {
+  0% {
+    transform: translateX(-33.33%);
+  }
+
+  100% {
+    transform: translateX(-66.66%);
```
olgalepisto: Love the animation!

my-react-portfolio/src/App.css
```diff
@@ -0,0 +1,176 @@
...
+} */
+
+/*media query for tablet*/
+@media screen and (min-width: 668px) and (max-width: 1024px) {
+  h1 {
+    font-size: 56px;
+  }
+
+  .projects-box {
+    grid-column: span 2;
```
olgalepisto: I see you have defined the span of the grid columns for multiple media queries, but I couldn't find anywhere that there would be a different grid-template-columns declaration for any media query (could also be that I just didn't spot them). I'm wondering if the responsiveness would be better if the amount of columns were different for desktop, tablet and mobile, as now I guess the media queries are just trying to adapt to the amount of columns in the desktop version?

olgalepisto (commented):
First of all, I'm really sorry this review is SUPER LATE.

Your code looks really neat and is easy to keep up with. I added a few comments mainly regarding the styling - I'm not sure if all of them are that helpful, as I also noticed that you're using some CSS modules which I know nothing about :D 
But great work overall, just some more styling and you're good to go!

my-react-portfolio/src/Components/FeaturedProjects/FeaturedProjects.jsx
olgalepisto: Okay apparently GitHub picked up nbsp from my comment so that's the mysterious space in there... xD

PR #310, opened 2023-10-15
Emmy Dieden's portfolio

emmydieden:
Due to lack of time there are still some details that are not in place. 

First of all: The structure of the components in the folders are right now a bit of a mess. Due to problems with deployment and broken paths I had to move the components around. By placing them in the current locations deployment worked and right now I don't dare moving them, as this would maybe interfere with the deployed version. Although, when there's time I will look over the folder structure. 

Duplicate: There's a duplicate of a component (ArticleCard.jsx). This should not be there and I will remove this, when I'm sure it won't interfere with the functionality.

Images: Can't get the right height for the images in tablet and desktop view. I find it tricky to keep the ration of the image and still display the right picture size. 

Wavy line: Have inserted this as a background image and therefor they won't join, which breaks the line up. Need to add the wavy line in a different way. 

Accessibility: The site scores 100% when checking with lighthouse. Cross-checking has been done in FireFox, Chrome and Edge. 

To view the deployed site: 
https://emmy-dieden-portfolio.netlify.app/


src/components2/projectCard/ProjectCard.jsx
```diff
@@ -0,0 +1,70 @@
...
+                  />
+                </div>
+              </div>
+            </div>
+          );
+        })}
+      </section>
+    </>
+  );
+};
```
jennylarsen: Looks nice and clean!

jennylarsen (commented):
Nice job with the project Emmy! Your page looks really nice and I love the color choice. You got a great score in Lighthouse. As you say there is only some minor styling details left to do! We both had the same issue with deploying and that we changed so many files (the node moduls?) so I can't see that much of the code. :( So yes I understand your issues with the components and that it is a bit messy. But I think that you should be proud of yourself and things that is not in place right now will be in the future, 100%.

PR #313, opened 2023-10-15
Portfolio Frida Lindskog

FridafridaL:
Finnished version

Link:
https://fridalindskog-portfolio.netlify.app/


andrkavr (approved):
I couldn't get an overview of your project because of the node modules having been uploaded and freezing my browser when I tried to check in pull requests.
All in all, I think you've done a great job with what you have finished, and you'll complete the rest with some more time. The structure is good and the code is easy to follow. Nicely done!
I have added the remarks I had below, though I believe that they are nothing you don't already know. 
I added a link to my footer design if that helps. It's not perfect, but maybe there's something there that's useable. 
https://github.com/andrkavr/Portfolio-Project/tree/main/src/components/footer

Smaller Screen
Entire page jumps from footer
Maybe add overflow: hidden?
Too much spacing between live demo and view the code
Squiggly line is cut off at the top


Medium screen
Setup of introduction is not complete. Unfortunately don't have any advice regarding that as I am struggling with it myself. 
Left part of border missing on images
Article images should be wider
Skills should be two columns

Large screens
Skills should be 4 columns
My words should be 2 columns
Introduction needs to be grouped and padding should be added

General remarks
Consider adding folders for your different sections and components so that each section/component and its corresponding css file are in one folder instead of all in the same. 
Consider adding border around profile picture so it won't blend in as much with the background.
LinkedIn icon link does not work.
In tech, spelling error: mob-programmin.
Consider changing to capital letter on both mob-programming and pair-programming to follow same convention as other listed tech.
Readability of Upcoming and More is low. Consider different background color or font-color.

PR #319, opened 2023-10-15
Project-portfolio

sansan-sakura:


src/styles/mixins.scss
```diff
@@ -0,0 +1,31 @@
+@mixin padding-reset {
```
elbines, line 1: i think the way you use folders and components and declare them in a structured way is really cool.

elbines (commented):
Your code is notably clean and well-structured. It appears that you have a system of declaring functions in specific folders and then reusing them throughout?

I must admit, my expertise isn't as advanced, so I occasionally find it challenging to review because I'm not always sure about certain elements. Even though the code is clean, I sometimes struggle to pinpoint where certain imports or declarations are made.

Nonetheless, I'm genuinely impressed by your site. The design is striking and I particularly appreciate the added features: the progress bar indicating how far the reader has scrolled, the language switch capability, and the "back to top" button. Everything comes together beautifully!

PR #320, opened 2023-10-15
Klaudia W

klaudiaW0410:
https://portfolio-klaudiaw.netlify.app/


Isaackamran (commented):
The first draft of the project looks very good, and has a lot of potential which I think with more time you can easily fix. I think you could/can "merge" the files in folder Footer together to keep everything together in the same file, but I suppose it's subjective if you want more components or not. Other than that, easy to read and follow. Great job Klaudia as always.

Isaackamran:
Also great performance when i'm using the lighthouse extension on firefox.

PR #332, opened 2023-10-16
Mirelcac Portfolio

mirelcac:
Still having deploy issues -managed to deploy, but page is blank


src/components/content/contact/contact.jsx
```diff
@@ -0,0 +1,34 @@
...
+                mainHeading="Let's talk"
+            />
+            <div className={style.innerBox}>
+                <img className={style.imgMe} src="/images/me.jpg" alt="Me" />
+                <div className={style.contactInfoBox}>
+                    <div className={style.contactInfo}>
+                        <p>Mirela Cacan</p>
+                        <p>+47 410 45 384</p>
+                        <p>mirelacacan@gmail.com</p>
+                    </div>
```
sansan-sakura: I think it is okay like this, but I think it might be nice to write like this. So when a user clicks, it will direct to mail or tel.
`<p><a href="mailto:someone@example.com">Send email</a></p>`
`<p><a href="tel:+4733378901">+47 333 78 901</a></p>`

src/components/content/contact/contact.jsx
```diff
@@ -0,0 +1,34 @@
...
+                    <div className={style.socialMedia}>
+                        <img src="./public/images/github.svg" alt="Github" />
+                        <img src="./public/images/linkedin.svg" alt="LinkedIn" />
```
sansan-sakura: `  LinkedIn`
You can write without public to update your images!!

src/components/content/contact/contact.module.css
```diff
@@ -0,0 +1,54 @@
+.contactBox {
+    display: flex;
+    width: 1280px;
```
sansan-sakura: I think if you write min-width or/and max-width, it might be easier to style for responsive. 
Max-width---> it doesn't go bigger than that.
min-width--> no smaller than that.

So if you give max-width for each sections, it doesn't grow too big:))

src/components/content/contact/contact.module.css
```diff
@@ -0,0 +1,54 @@
...
+    padding: 128px 433px 128px 434px;
```
sansan-sakura: I did the same when I started this project, copying the code from Figma, but I got into trouble. So I found it easier to just define top/bottom paddings for bigger elements that will grow by screens.

src/components/reusable/mainheader/mainheader.jsx
```diff
@@ -0,0 +1,9 @@
+//import {useState, useEffect} from "react"
+import style from './mainheader.module.css';
+
+export const MainHeader = ({ mainHeading, className }) => {
+    console.log(className)
+    return (
+        <h1 className={`${style.SectionH1} ${className}`}>{mainHeading}</h1>
+    )
```
sansan-sakura: I think in this case (you are using modules), I think you want to do this.
`className={`${style.SectionH1} ${style[className]}`}`

src/components/content/footer/footer.jsx
```diff
@@ -0,0 +1,15 @@
...
+export const Footer = () => {
+
+    return (
+        <div className={style.footerBand}>
+            <p className={style.footerText}>
+                Mirela Cacan · Web Development · Mirela Cacan · Web Development · Mirela Cacan · Web Development · Mirela Cacan · Web Development · Mirela Cacan · Web Development · Mirela Cacan · Web Development ·
+            </p>
+        </div>
+    )
+}
```
sansan-sakura: I think you can repeat the sentence using Array.from({length:10},(_,i)=><p></p>}) Something like this. You can search more about this method because I think this is very useful!!  This creates the same element * 10(length). 
So your footer p is long enough!!

sansan-sakura (commented):
I think you tried hard to work on this project and do a good job.

Here are few tips I think I can give you.
1. Try not to use width, but max-width/min-width. I know there are situations that you need width, but when you use that , it needs more care when it come to responsiveness. It is also important to remember that giving padding might work well, as width can grow. 
2.But Padding can cause problems, if you give too many paddings. -> then try max-width/min-width.
3. I found that working with React, you should give #root div 100vw to expand full screen.
-> If there is any questions, let me know :))

PR #336, opened 2023-10-17
Maria's portfolio

mariateresepettersson:
https://mariateresepettersson.netlify.app/


Cocofnas (commented):
Hi, the github only loads 3000 files and i really couldn´t find the project portfoliofolder. But som comments, I like that you displayed only the projects you are most proud of, smart move and I'm impressed that you managed to toggle the view and github/linkedin buttons. Lastly your footer is great, mine just restarts. So great, just wished I could have viewed the code

src/components/button/Button.jsx
```diff
@@ -0,0 +1,23 @@
+import './button.css'
```
Cocofnas, line 1: smart solution with having a button component!

src/components/button/button.css
```diff
@@ -0,0 +1,47 @@
+button {
+    border-radius: var(--corner-radius);
```
Cocofnas, line 2: I don´t think I've come across this before, what and why is the use of 'var' smart?

src/components/typography/Heading.jsx
```diff
@@ -0,0 +1,17 @@
...
+};
+
+//Define default and customizable props
+Heading.defaultProps = {
+    level: 1, // Default to h1
+    className: '',
+};
+
+
+{/* THIS IS HOW YOU USE IT
```
Cocofnas, line 15: great feat to control it from a top level instead of every little component!

README.md
```diff
@@ -71,13 +30,15 @@ Accessibility has been a priority, following guidelines to make my website usabl
...
+Through this project, I, Maria, have learned invaluable lessons in React development. I now understand the power of components and props in creating a modular and organized codebase. React hooks, especially useState, have allowed me to manage state and interactivity effectively. Integrating web APIs taught me the importance of data fetching and dynamic content display, keeping my portfolio up-to-date effortlessly. Additionally, focusing on accessibility has made me more aware of creating inclusive user experiences. Overall, this project has honed my skills in clean coding practices and emphasized the significance of continuous testing and deployment for a successful web application. The experience has been enriching, and I am excited to apply these lessons in future projects to further enhance my coding journey.
+
+## Deployed Site
+
+https://mariateresepettersson.netlify.app/
```
Cocofnas, line 37: it looks good,but there is something goung on with the mocie project, the buttons overlap the techniques used :/

Cocofnas (commented):
It looks so good and I can see that you put a lot of work into it!!!

src/components/button/button.css
mariateresepettersson: It is used to create global styles that can be defined in and then you use semantics to recall that code.

README.md
mariateresepettersson: Great you saw this, finally I think I've fixed it :)

PR #339, opened 2023-10-20
New Repo, copied from old

Isaackamran:


Isaackamran:
Code review by Frida Lindskog:

Nice work Isaac! You’re on your way to make a nice portfolio!!
I noticed some things:
Your picture: if you take away the fixed height (height: 150px) it won’t be stretched.
When clicking on live demo and view code I’m linked to my own GitHub and netlify. You have to link your public GitHub link to make the user come to your site. Here is your:
https://github.com/Isaackamran
And for netlify it will be good if you link to each project, like the finnished deploy link to each project.
A tip is to do a json-file for the articles aswell as you did for the projects. In that way you don’t have to repeat in the section.jsx
And also do some responsive design right?
Otherwise a really nice structure with the folders, makes it easy to find each section and the css for each section.
Great work Isaac yeey yeey yeey!

PR #349, opened 2024-03-27
my portfolio 

Josephineoderland:
## Netlify link
https://rococo-mousse-b4a128.netlify.app/


HIPPIEKICK (changes requested):
⭐️
- Skills section looks perfect!
- So do the hovered buttons ⭐️
- Nice that the arrow is clickable, makes total sense. Nice scroll transform
- Nice usage of React state

**Needs another take**

- Overall, you need to have another look at distances and sizes of your elements. The margins/paddings are not the same as in the design. Distance on the sides in mobile should be 16px. I suggest removing the margin on each element and just change your 5px padding on the .page to 16px
- Tags should have more padding
- Buttons’ text should not go on two lines, it’s OK to decrease font-size to make sure this doesn’t happen
- Skills section should be left aligned in tablet and up + have more space between columns
- Tech section is too cramped in 1281px width. All sections should have the same distance to the sides except for the Let’s talk section

**Left todo**
- Date tag + button for article module
- Remove box-shadow on project images

**Tips**
- For images and especially icons, it’s always best to use svg to get them as crisp as possible. They can easily look pixly otherwise
- object-fit: cover; on images to get rid of distortion
- `<br>` is deprecated, use flexbox or something instead

Josephineoderland:
commited!
the Needs another take and Left todo

HIPPIEKICK:
I think maybe your Netlify is not reflecting the changes then? 👀 https://rococo-mousse-b4a128.netlify.app/

Josephineoderland:
> I think maybe your Netlify is not reflecting the changes then? 👀 https://rococo-mousse-b4a128.netlify.app/. fixt!

HIPPIEKICK (approved):

PR #354, opened 2024-03-30
Linda's React Portfolio

linda-f:
## Netlify link
https://linda-f-project-portfolio.netlify.app/

## Collaborators
[linda-f]


HIPPIEKICK (changes requested):
⭐️
Nice animation of the arrow! And well done paying attention to distances. Amazing work overall, but I need to put Changes Requested mostly because of the missing article section.

**Have another look**
- I know the figma file says font-weight 600 for the Tech words, but I think it looks more like normal font-weight, so feel free to change that (it does look better with normal, doesn’t it?)
- Skills section should have two columns in tablet
- Profile text is a bit too narrow, it should align with the other sections

**Todo**
- Article section

**Tip**
Apps should always have only one h1, it’s like the title of a book, even though they might be equally sized.

HIPPIEKICK (approved):

PR #355, opened 2024-03-30
Portfolio Viktoria O.

viktoria-olo:
## Netlify link
https://remarkable-shortbread-2696c2.netlify.app/

## Collaborators
Individual project


HIPPIEKICK (changes requested):
⭐️
Nice animation of the arrow! I like the color choices as well, looks aligned and nice. Good job overall!

**Needs another take**
- "Hi, I'm Viktoria"-text should be 18px in mobile
- Between Web Developer and the paragraph it should be 32px
- Side padding/margin in mobile should be 16px, you have 16px + 8px. Remove the margin of the body
- I know the figma file says font-weight 600 for the Tech words, but I think it looks more like normal font-weight, so feel free to change that (it does look better with normal, doesn’t it?)
- Make sure to double check all distances, e.g. top/bottom margin/padding of all the section headlines.
- Skills section font-size should be 18px
- Desktop especially: Make sure all sections align (having the same padding/margin on the sides)
- Broken squiggly line
- Thicker image borders in desktop

**Todo**
- Article section (OK to save for later)
- Tablet layout:
  - Top section is not the right layout
  - Tech section needs more padding
  - Featured projects should be image to the left, text + button to the right
  - Skills section should be two columns

HIPPIEKICK:
Looks much better, great job ⭐

HIPPIEKICK (approved):

PR #362, opened 2024-04-01
Wen's portfolio website

wwenzz:
## Netlify link
https://wen-portfolio.netlify.app/

## Collaborators
[wwenzz]


HIPPIEKICK (approved):
Hi Wen! Since you did not follow our design, I can't review the biggest part of this assignment which was to follow our design as closely as possible. I'm curious - in our instructions we said that if you don't want to follow our design, you have the possibility to present your own design to us - why didn't you? 👀 

Anyway, to me (without a design eye) it looks nice, and you have clearly put a lot of effort into this project so I will not ask you to redo everything. What I would like you to do though, is to have a look at [this resource](https://uxplanet.org/16-ui-design-tips-ba2e7524d203) and see if you find anything you can improve. I would consider aligning your elements (so the sections have the same width) and have another look at how your app looks on wide screens (very long lines of texts are hard to read).

PR #365, opened 2024-04-01
Portfolio - Pernilla

pernillasterner:
## Netlify link
https://pernillasterner.netlify.app/


HIPPIEKICK (changes requested):
⭐️
Really nice with a Show More button! You have really paid attention to the design and it shows 🙌  Some relatively small things that are not according to the design:

- Margin bottom on the h2s should be 64px in mobile and tablet
- Tech words should be left aligned
- h2s should have 66px line-height
- Have a look at the different distances within a project card
- Pick between margin and padding for the project tags. 2px 6px is enough, no need for double. Maybe consider decreasing the amount of tags or making the images higher so that the images align better with the text.
- h3s in Skills section should not be smaller than the list items
- Be aware that the arrow comes very close to the text on tablet (745px)
- The project section doesn't have the right layout in tablet. It should be image to the left and text to the right.
- Skills columns should be left aligned in tablet and up
- You're contact info is a bit on the big side in tablet
- Make sure all sections align

pernillasterner:
Hi @HIPPIEKICK , do you think the changes requested are approved? :)

HIPPIEKICK (approved):

PR #367, opened 2024-04-01
Sofie Ferrari Strahl -  portfolio

SofieFerrari:
https://sofieferraristrahl.netlify.app/


HIPPIEKICK:
Hej Sofie! Please remove your node modules from GitHub so that I can review your project 😊

HIPPIEKICK (changes requested):
⭐ 
Good job! I can see you paid attention to a lot of details in the design, and you're on good way here - but there are still some things that I would like you to have another look at.

**Needs another take**
- First I would like you to look at the responsiveness. In 320px width it's broken because of the Skills section and your email address being to wide
- Remove the margin from the body
- If you don't want to create the scrolling banner, make sure the text is not on multiple lines
- All clickable elements should have cursor: pointer;
- Your profile image is not completely round in mobile

I would also like you to have another look at the design and check the following for each section and device width:

PS. Upcoming is misspelled.

AntonellaMorittu (approved):

PR #368, opened 2024-04-01
W8 - Portfolio-Maria

Maria-Manuela:
## Netlify link
https://bmmportfolio.netlify.app/


HIPPIEKICK (changes requested):
Hi! Please make sure you're following the design and that your project is responsive (have a look at 320px width). Here are some questions to start from:

HIPPIEKICK:
Keep up the good work, you're getting there 👏 Let us know if you need any assistance.

NellaMori (changes requested):
Hi MM, I can see you applied some changes, but the font sizes are still not exactly the same ones of the design. Please copy all the css rules in terms of sizes, margins and paddings, this is part of the exercise and assignment. Follow the design layout by the pixel :)

NellaMori (approved):

PR #369, opened 2024-04-01
Portfolio project

kathinka:
## Collaborators
solo


my-react-portfolio/public/assets/Favicon/android-chrome-512x512.png
ohitsnathalie, line 1: omg I love your favicon!!

my-react-portfolio/src/Components/Email.jsx
```diff
@@ -0,0 +1,15 @@
+
+export const Email = () => {
+  const part1 = "kathinkasewell";
+  const part2 = String.fromCharCode(Math.pow(2, 6));
+  const part4 = "gmail.com";
+  const email = `${part1}${part2}${part4}`;
```
ohitsnathalie, line 6: Clever solution to deconstruct it so you can't easily get spam mails!

my-react-portfolio/src/Components/Image.jsx
```diff
@@ -0,0 +1,21 @@
...
+const profileImage = "../assets/kat.webp";
+const style ="profile-image";
+
+return (
+  <picture className={style}>
+<source srcSet={profileImage}
+  alt="Kathinka Sewell" type="image/webp" />
+<source srcSet={profileImageBackUp}
+  alt="Kathinka Sewell" type="image/png" />
+<img className="profile-image wide"
```
ohitsnathalie: May I ask why you have two different images? 🤔  I saw mobile has one version and tablet/desktop the other.
I like that you used webp. It's supposed to be the better choice for image format 😄

my-react-portfolio/src/Components/Introduction.jsx
```diff
@@ -0,0 +1,38 @@
+import "./introduction.css";
+import { Image } from "./Image";
+
+export const Introduction = () => {
+  const profileImage = "../assets/kat.webp";
+  const profileImageBackUp = "../assets/kat.webp";
+  const arrow = "../assets/Arrow.svg";
+  const copy = "Frontend Developer with a background in Graphic design, UX/UI, Art Direction and with experience leading agile development projects as a scrum master. Passionate about creating web applications that not only looks good, but also are user-friendly and accessible. Always eager to learn new technologies and improve her skills."
+  return (
+    <section className="center-wrap">
```
ohitsnathalie: It seems like you don't use this "center-wrap".  You have your main "wrapper". 
To save some space in the DOM you can just use the fragment <> </>

my-react-portfolio/src/Components/Image.jsx
```diff
@@ -0,0 +1,21 @@
...
+  src={profileImageBackUp}
+  alt="Kathinka Sewell"
+/>
+</picture>
```
ohitsnathalie: Since you want to have several images in this component, you could write it like that:
`export const imageHeader = () => {
  return (
    Kathinka Sewell
  )
}
  export const imageContact = () => {
    return (
      Kathinka Sewell
    )
  }`
  
 maybe this is an approach you want to do.
 EDIT: Formatting code is not really possible ☹️

my-react-portfolio/src/Components/Introduction.jsx
```diff
@@ -0,0 +1,38 @@
...
+      <main className="wrapper">
+        <section className="title">
+          <picture className="profile-image mobile">
+            <source srcSet={profileImage}
+              alt="Kathinka Sewell" type="image/webp" />
+            <source srcSet={profileImageBackUp}
+              alt="Kathinka Sewell" type="image/png" />
+            <img className="profile-image mobile"
+              src={profileImageBackUp}
+              alt="Kathinka Sewell"
```
ohitsnathalie: I already commented on how you could save some lines for the image section in the Image.jsx

my-react-portfolio/src/Components/Introduction.jsx
```diff
@@ -0,0 +1,38 @@
...
+            />
+          </picture>
+          <h1>Hi, I&apos;m Kathinka Sewell</h1>
+          <h2>Frontend Developer</h2>
+        </section>
+        <section className="introduction wide">
+          <Image/>
+          <article className="intro">
+            <p>
+              {copy}
```
ohitsnathalie: Here you could have a component where you would link your copy to, instead of writing the copy in the top as a const. 

So you could just change for example the Copy.jsx and in the code you would write it as `<Copy />`

my-react-portfolio/src/App.css
```diff
@@ -0,0 +1,47 @@
...
+	margin: 0;
+	padding: 0;
+	background-color: var(--bg-color-bright);
+}
+
+
+body {
+	color: var(--text-color);
+	font-family: hind-regular, sans-serif;
+	margin: 0 auto;
```
ohitsnathalie: You already had body margin defined. The website has a x-overflow on mobile.

my-react-portfolio/src/Components/Projects.jsx
```diff
@@ -0,0 +1,111 @@
+import { useState, useEffect } from "react";
+import { PortfolioImages } from "./PortfolioImages";
+import { fetchProjects } from "./Utility";
+import "./Projects.css";
+
+// Projects component fetches data from the GitHub API and displays the projects in the portfolio.
+export const Projects = () => {
+  const [projects, setProjectList] = useState([]);
```
ohitsnathalie: I believe best practice is to name it `[projects, setProjects]` or `[projectList, setProjectList]`

my-react-portfolio/src/Components/Projects.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+  const currentDate = new Date();
+  const currentYear = currentDate.getFullYear();
+  // lets keep the portfolio fresh! - returns a list of repositories created since the beginning of the current year.
+  const url = `https://api.github.com/users/kathinka/repos?per_page=100&since=${currentYear}-01-01T00:00:00Z`;
```
ohitsnathalie: Oh that's a nice idea. Good to keep your newer projects visible.

my-react-portfolio/src/Components/Projects.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const projects = await fetchProjects(url);
+        setProjectList(projects);
+      } catch (error) {
+        <alert>Oh no! I failed to fetch any projects😅 - please try again🙏🏽</alert>
```
ohitsnathalie: haha funny! 😄

my-react-portfolio/src/Components/Projects.jsx
```diff
@@ -0,0 +1,111 @@
...
+    updated_at,
+    pushed_at,
+    full_name,
+    homepage,
+    topics,
+    stargazers_count,
+  }));
+
+  //filtering out the projects that have no stars, as for now I have chosen that as my filter for what goes into the portfolio. later I will potentiall use a different filter, ect. only show projects that have been updated in the last x months/years etc.
+  const starredProjects = projectObject.filter((project) => project.stargazers_count > 0);
```
ohitsnathalie: Nice idea, I went with the approach to show only the projects that has a "homepage" aka deployed link.

my-react-portfolio/src/Components/Projects.jsx
ohitsnathalie, line 1: You could have broken down the Projects component into smaller pieces. Then you would have more little code snippet instead of one large code pile.

my-react-portfolio/src/Components/Projects.css
```diff
@@ -0,0 +1,220 @@
...
+}
+
+button {
+  cursor: pointer;
+  height: 48px;
+  all: unset;
+}
+
+button.github {
+  background-image: url(../assets/Buttons/View-Code-button.svg);
```
ohitsnathalie: This is actually a clever solution without struggling with too much CSS. Amazing 🌟

my-react-portfolio/src/index.css
ohitsnathalie, line 1: Next time you could try just to use either index.css or App.css, because both of them have :root in it and you could keep it more clear.

ohitsnathalie (commented):
Congratulations on your portfolio🎉 Great job, Kathinka!
It has the look and feel of the Technigo design. My general comment here is, that you could make your code a bit dryer. It seems like you took a longer route to get to the same destination 🗺️ 
Try to think in the beginning what can you break down in smaller bites and pieces. But because you took the unconventional route, you found some really clever tricks, like button image 🤯 
Also, I love to encourage people to write their README. Because I just like to read what people thought during their process.

With that said, I made you some comments below. Maybe you agree, or you can enlighten me with your way of thinking, because it's always interesting to see how other approach a task like this. If you have questions or if something is not clear, just slide into my DM 😉

my-react-portfolio/src/Components/Introduction.jsx
kathinka: good idea - it would require more code, but will probably be even easier to maintain in the long run 🤓

my-react-portfolio/src/Components/Email.jsx
kathinka: let's hope it works 🤞🏽

my-react-portfolio/src/index.css
kathinka: I'll just remove this one - it was not supposed to be included in the final version (sloppy cleanup 😅)

my-react-portfolio/src/Components/Image.jsx
kathinka: Its because I am lazy - Instead of making elaborate grids to move the image, I just have two and hide the one I dont need for  the different layouts. (so one for mobile, and another one for pad and dt) - maybe I will look into how I can solve it in a less lazy way in the future :)

my-react-portfolio/src/Components/Projects.jsx
kathinka: it is - i will fix it. (oh the shame!😅)

my-react-portfolio/src/Components/Projects.jsx
kathinka: That is clever - I might have to adjust my approach!

my-react-portfolio/src/Components/Projects.jsx
kathinka: Its important to laugh when the world, or you code breaks.

my-react-portfolio/src/App.css
kathinka: yup - it was sloppy. a bit rushed this last "versjon" of this project. thanks for pointing it out!

my-react-portfolio/src/Components/Introduction.jsx
kathinka: Actually I do - but it only goes into play on pad sizes. without it it would fail to position center. and since  wrapper is a grid, i cant use flexbox for positioning as i do with center-wrap :)

HIPPIEKICK (changes requested):
Hi! Please make sure you're following the design and that your project is responsive (have a look at 320px width). Here are some questions to start from:

NellaMori (approved):

PR #373, opened 2024-04-02
technigo-w8-project-portfolio

schouenkes:
## Netlify link
https://technigo-w8-project-portfolio.netlify.app/


HIPPIEKICK (changes requested):
Hi! You're very close to the design, but I would like you to have another look at the following questions, on all device widths:

- Are all sizes and distances (margin, padding) according to the design? Are the sections aligned where they should be?

my-portfolio/src/sections/Contact.jsx
```diff
@@ -0,0 +1,26 @@
...
+            <img src="icons/github-tag.svg" alt="GitHub icon" />
+          </a>
+          <a href="www.linkedin.com/in/maja-schouenke-0a09842a7" target="_blank" rel="noreferrer noopener">
+            <img src="icons/linkedIn-icon.svg" alt="LinkedIn icon" />
+          </a>
+        </div>
+      </section>
+    </div>
+  );
+};
```
Paulajungaker, line 29: When I try to open the linkedIn button I get "page not found" so have a look at this link if its working properly.

my-portfolio/src/sections/Tech.css
```diff
@@ -0,0 +1,26 @@
...
+  font-weight: var(--font-weight-700);
+  text-align: center;
+}
+
+.tech-container p {
+  color: #fff;
+  font-family: var(--text-font);
+  font-size: var(--font-size-24);
+  font-weight: var(--font-weight-600);
+}
```
Paulajungaker, line 86: To make it a bit more like the design you should probably add some padding to the left and right for the Tech part if the screen is wider than 375px.

my-portfolio/src/sections/FeaturedProjects.css
```diff
@@ -0,0 +1,142 @@
...
+}
+
+.github-button:hover {
+  background-color: var(--primary-color);
+  color: #fff;
+}
+
+.github-button:hover img {
+  filter: invert(100%) brightness(2);
+}
```
Paulajungaker, line 60: The Featured projects images take up the full width of the screen. It looks very good when you view on phone but on larger devices the design should be a bit different.

Paulajungaker (commented):
Your sections and components code looks great and clean! I like how you have structured the project. There are some issues in the CSS where the I think you should add some padding to the Tech-section if the screen is larger than 375px to make it more like the design we should follow. Also the link to linkedIn in your Contact-section is not working for me so might need to check that. The images in the FeaturedProjects-section takes up the whole width of the page. It looks great in a mobile view but on larger screens it does not look like the design. I like the colors you have used and your portfolio looks very cool. Great job Maja with the portfolio:)

HIPPIEKICK:
It looks more aligned now, good job! But please have a look again at how it looks in 320px width. It's a bit too cramped 👀

schouenkes:
I hope my solution fixed it properly :)

HIPPIEKICK (approved):

PR #374, opened 2024-04-02
Project-portfolio

Paulajungaker:
## Netlify link
https://amazing-pegasus-a92fd6.netlify.app/


HIPPIEKICK (changes requested):
If you want to keep the dark/light mode styling, you need to account for changing the background color too. Right now I get white text on light pink background, which is not working accessibility wise. Apart from that, please have another look at the design and go through the following questions on all sizes:

- Are all sizes and distances (margin, padding) according to the design? Are the sections aligned where they should be?

my-react-portfolio/src/styling/Skills.css
```diff
@@ -0,0 +1,107 @@
+.skillsContainer {
+  background-color: #5f5d9c;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  color: white;
+  font-family: Montserrat;
```
viktoria-olo, line 7: Since there are a lot of instances where Montserrat is specified, an option could be to set all h1, h2 etc. to Montserrat in a css file that covers all components, like App.css or index.css

my-react-portfolio/src/projects.json
viktoria-olo, line 1: Nice job with the json! Very convenient to add all the needed info in one place and gives you a lot of control over what is displayed. Makes it easy to add endpoints as well, in case you want some more info on your page.

viktoria-olo (commented):
It looks really good! I like the colour scheme, the code is easy to read and follow along with. Great job!

AntonellaMorittu (approved):

PR #381, opened 2024-05-10
Portfolio

fridaforser:
## Netlify link
https://fridaforser.netlify.app/


NellaMori (changes requested):
Hey Frida, your portfolio looks nice, although is not as close as possible to the design. It was a requirement to pass. 
Check also:
- Accessibility: You should have a score of at least 95 in
- How to write good code: format your code + format to 2-spaces tab all your files
Thank you 🤝

PR #388, opened 2024-11-03
Portfolio-Emelie-Nyberg

EmelieNyberg:
## Netlify link
https://emelie-nyberg-portfolio.netlify.app/

Not 100% pleased with the result. If I had more time I would make som more reusable components and better responsivness on mobile devices.


JennieDalgren (approved):
Great job with this project!
You've met all requirements. However the a11y score on Lighthouse is a bit low, but I will doublecheck this with the scoring in Figma and that design, because it's on the designer, not on you!

Keep up the good work!

PR #398, opened 2024-11-04
Portfolio

gittebe:
https://portfolio-gitte.netlify.app/


src/components/MediaButtonContainer/MediaButtonContainer.jsx
HeleneWestrin, line 1: I find the name MediaButton and MediaButtonContainer a bit unclear. To make it more readable I would suggest renaming it to SocialMediaButton, even though that's longer.

src/App.jsx
HeleneWestrin, line 1: Great structure, so easy on the eyes and mind 😍

src/sections/SkillsSection/SkillsSection.jsx
HeleneWestrin, line 1: I like your solution for how to enter the data here! I went with a JSON-file, but it feels a bit excessive when there is not that much data, so this was cleaner I think.

src/components/ProjectCard/ProjectCard.jsx
HeleneWestrin, line 1: I was expecting to see the map method in the ProjectsSection.jsx and not here in the ProjectCard component. If you ever wanted to make this a reusable component I would suggest moving that logic outside the ProjectCard.jsx and pass the data in via props instead.

HeleneWestrin: Also noticed a console.log still present in the code. Maybe remove that if you don't need it anymore.

src/App.css
HeleneWestrin, line 1: It says this file is empty, so maybe you should remove it?

src/ui/Grid/Grid.css
```diff
@@ -0,0 +1,21 @@
...
+
+@media (min-width: 667px) {
+  .grid-section {
+    grid-template-columns: repeat(8, 1fr);
+  }
+}
+
+@media(min-width: 1024px) {
+  .grid-section {
+    grid-template-columns: (12, 1fr);
```
HeleneWestrin: The repeat function is missing in `grid-template-columns: (12, 1fr);`, so it never gets applied on desktop.

src/sections/BioSection/BioSection.css
HeleneWestrin, line 1: You seem to be mixing CSS Grid and FlexBox attributes here in a way that I am not sure actually works. In the browser it looks like `.bio-header { grid-column: span 4; }` is not getting applied because the parent is Flex for instance.

HeleneWestrin (commented):
Fantastic job with the portfolio, Gitte! 🌟  The code was really easy to follow and read. I left a couple of comments on potential improvements, but overall it looks really great.

**I can see some things that differ from the design:**

1. The thing that I noticed first is that you didn't add a max-width to the grid. In the design it looks like all sections except projects and the articles follow the grid. Setting `max-width: 1084px;` and `margin: 0 auto;` to .grid-section should do the trick. You might have to adjust some other stuff though because of the change.
2. HeaderTwo has the wrong size on desktop.
3. The body font-size for the tech stack is too small and it doesn't seem to be a max-width here either. Looking at the design it seems it should span 8 columns of 12 on desktop.
5. The footer has the wrong layout on desktop.  I think it might be because you placed a section inside the .grid-section instead of having the two divs with the footer content as direct children of .grid-section.
6. Not related to the design, but I think it would be nice to at least add `cursor: pointer; ` to the buttons so that they look clickable when hovering. Some sort of hover state would be nice too.

It was inspiring to look through you code and see how you approached different sections. I might just steal some of it for future projects 👏

JennieDalgren (approved):
Great job with the portfolio!
You are meeting all the requirements.
Good structure of your files too.

Stay golden! ⭐️

PR #401, opened 2024-11-13
Fannys Portfolio 

Fannyhenriques:
## Netlify link
https://fannys-portfolio.netlify.app/


HIPPIEKICK (changes requested):
Good job! I'd just like to ask you to do another take on the buttons:
- hover effect
- correct font
- remove border
- change them to `<a>` tags instead of using button with onclick

Apart from that, it looks super ⭐

Fannyhenriques:
Thank you so much for the feedback, @HIPPIEKICK! ☺️ I’ve updated the buttons as suggested. However, I’m a bit unsure about what you meant by removing the border. The buttons with borders (the "View More" buttons) are styled that way to match the Figma design. Removing the borders would make the site differ from the original design. Could you clarify if you meant something else? 😊

HIPPIEKICK:
> Thank you so much for the feedback, @HIPPIEKICK! ☺️ I’ve updated the buttons as suggested. However, I’m a bit unsure about what you meant by removing the border. The buttons with borders (the "View More" buttons) are styled that way to match the Figma design. Removing the borders would make the site differ from the original design. Could you clarify if you meant something else? 😊

I think you solved it by changing to a-tags! The button-tags has a default border, but a-tags do not 👍

HIPPIEKICK (approved):

PR #402, opened 2024-11-18
Maryyy_ux_Portfolio

Maryyy-ux:


HIPPIEKICK (changes requested):
Hi Maria!
You're off to a good start with your portfolio. However, there are still things left before it's following the design. Pay extra attention to spacings, fonts, text-alignment and sizes.

PR #403, opened 2024-12-06
Project-portfolio 

Heleneabrahamsson:
https://helenes-portfolio.netlify.app/


HIPPIEKICK (changes requested):
Remove your node modules so that we can do a proper review of your code, as well as remove any dark/light mode styling you have as I think that's what's causing your portfolio to look broken. After that I can do a more thorough review.

Skärmavbild 2024-12-09 kl  11 37 06

HIPPIEKICK (dismissed):
You're off to a good start Helene, but you still need to work on the design a bit. Make sure it looks good on all screen widths (and get rid of the side scroll) and take another look at the design. Be as thorough as you can and go through spacings and such.

JennieDalgren (approved):
🎯

JennieDalgren:
Code wise you are approved! - just remember to change the content to be yours. It says Sara in some places and the photo in the end of the site is the placeholder image.

